PR #1, opened 2020-06-29
add style check to workflow

haoyxc:
Just added format.yml so code can look nice!!


haoyxc:
@googlebot I signed it!

EatingW:
@googlebot I fixed it.

PR #3, opened 2020-06-30
Create a guava graph to store the data from proto graph

haoyxc:
Read input files, convert proto nodes to GraphNodes (self-defined object), add GraphNodes to Guava Graph, and apply list of mutations (from input file) to the Guava Graph.


EatingW:
> It looks like .settings/ and target/ are still present in this PR. Also, seems like .classpath and .project are also IDE specific and does not need to be committed. Could you remove those from this PR?

Actually looking more closely, the target/ changes are all deletions. Let's move these to a separate PR. This PR should only contain project file related changes.

haoyxc:
> > It looks like .settings/ and target/ are still present in this PR. Also, seems like .classpath and .project are also IDE specific and does not need to be committed. Could you remove those from this PR?
> 
> Actually looking more closely, the target/ changes are all deletions. Let's move these to a separate PR. This PR should only contain project file related changes.

Sounds good, I created a branch off FileRead addressing that. Will merge master into it (so there aren't so many commits) when this PR is merged into master!

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -25,16 +35,121 @@
...
+
+    // GRAPH Data structures:
+    // Create an undirected graph data structure to store the information, and
+    // map each node name in the graph to the GraphNode objects. This is the graph & map
+    // we update with mutations
+    MutableGraph<GraphNode> graph = GraphBuilder.directed().build();
+    HashMap<String, GraphNode> graphNodesMap = new HashMap<>();
+
+    // Create a graph (useable form) from proto data structures
+    for (String nodeName : protoNodesMap.keySet()) {
```
EatingW, line 58: I think this for loop could live in its own function. You should also write some unit tests for this function:
 - create a protoNodesMap
 - call this function
 - assert properties about the graph, such as nodes in the graph and edges in the graph

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -25,16 +35,121 @@
...
+        graph.putEdge(graphNode, protoNodeToGraphNode(protoNodesMap.get(child)));
+      }
+    }
+
+    // Parse the contents of mutation.txt into a list of mutations
+    List<Mutation> mutList =
+        MutationList.parseFrom(getServletContext().getResourceAsStream("/WEB-INF/mutations.txt"))
+            .getMutationList();
+
+    for (Mutation mut : mutList) {
```
EatingW, line 80: Also pull mutating the graph into a separate function (the function should only execute one step of the mutation, call this function using the for loop), with its own unit tests.

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -25,16 +35,121 @@
...
+      // Convert thisNode into a graph node that may store additional information
+      GraphNode graphNode = protoNodeToGraphNode(thisNode);
+
+      // Update graph data structures to include the node
+      graph.addNode(graphNode);
+      graphNodesMap.put(nodeName, graphNode);
+
+      // Add dependency edges to the graph
+      for (String child : thisNode.getChildrenList()) {
+        graph.putEdge(graphNode, protoNodeToGraphNode(protoNodesMap.get(child)));
```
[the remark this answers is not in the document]
IshaniSanturkar: We were relying on [this](https://guava.dev/releases/23.0/api/docs/com/google/common/graph/MutableGraph.html#putEdge-N-N-) implementation dependent behavior that if the node exists an edge is added but if it isn't both the node and an edge are added. I added an extra check for this case, but I'm not sure how we would end up with duplicate nodes?

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -25,16 +35,121 @@
...
+
+      // Getting the corresponding graph nodes from the graph map
+      GraphNode startNode = graphNodesMap.get(startName);
+      GraphNode endNode = graphNodesMap.get(endName);
+
+      switch (mut.getType()) {
+        case ADD_NODE:
+          // Create a new node with the given name and add it to the graph and the map
+          Node newNode = Node.newBuilder().setName(startName).build();
+          GraphNode newGraphNode = protoNodeToGraphNode(newNode);
```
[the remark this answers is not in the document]
IshaniSanturkar: Parsing the .proto file gives us a Graph object which contains a Map from string name to Node. We created a separate graph node because we thought that in the future we might add attributes like changedFromPrev graph (for color-coding the diffs) to the node but we couldn't modify the proto to get that in the Node object.

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -25,16 +35,121 @@
...
+          graph.addNode(newGraphNode);
+          graphNodesMap.put(startName, newGraphNode);
+          break;
+        case ADD_EDGE:
+          if (startNode != null && endNode != null) { // Check nodes exist before adding an edge
```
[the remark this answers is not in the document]
IshaniSanturkar: We'll incorporate this in the next PR which handles server-client communication :)

src/main/java/com/google/sps/servlets/DataServlet.java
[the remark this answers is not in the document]
IshaniSanturkar: I think I see what you mean, but the graph needs nodes to be unique so it is probably using the equals method of graph nodes to prevent duplicates and these two objects would be equal?

src/main/java/com/google/sps/servlets/DataServlet.java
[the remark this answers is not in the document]
EatingW: What I meant was why not just do something like `newGraphNode = GraphNode.create(...)` here, but I guess converting it to a proto Node first handles the default values for token_list and metadata :)

src/main/java/com/google/sps/servlets/DataServlet.java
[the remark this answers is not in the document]
EatingW: Ah, I see. We can make sure that this is the case with some unit tests ;)

src/main/java/com/google/sps/servlets/DataServlet.java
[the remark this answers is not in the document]
IshaniSanturkar: Yup exactly :)

src/main/java/com/google/sps/servlets/DataServlet.java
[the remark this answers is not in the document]
IshaniSanturkar: Yup I did this

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -126,30 +119,99 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
    * @param thisNode the input data Node object
    * @return a useful node used to construct the Guava Graph
    */
-  private GraphNode protoNodeToGraphNode(Node thisNode) {
-    return GraphNode.create(thisNode.getName(), thisNode.getTokenList(), thisNode.getMetadata());
+  public GraphNode protoNodeToGraphNode(Node thisNode) {
```
EatingW, line 122: What were the reasons behind this change? I thought that the previous version should have worked well.

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -126,30 +119,99 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+          return false;
+        }
+        // Create a new node with the given name and add it to the graph and the map
+        GraphNode newGraphNode =
+            GraphNode.create(startName, new ArrayList<>(), Struct.newBuilder().build());
+        graph.addNode(newGraphNode);
+        graphNodesMap.put(startName, newGraphNode);
+        break;
+      case ADD_EDGE:
+        if (startNode != null && endNode != null) { // Check nodes exist before adding an edge
```
EatingW, line 160: In here you have a lot of patterns like
```
case A:
   if (condition) {
         do something
         // expect to return true at the end of the function.
   } else {
         return false;
   }
```
which I think would be more easy to follow if you did:
```
case A:
    if (! condition) {
         return false;
    }
    do something
```

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -126,30 +119,99 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+        break;
+      case DELETE_EDGE:
+        if (startNode != null && endNode != null) { // Check nodes exist before removing edge
+          graph.removeEdge(startNode, endNode);
+        } else {
+          return false;
+        }
+        break;
+      case CHANGE_TOKEN:
+        if (startNode != null) {
```
EatingW, line 182: Here you'll be returning true if startNode is null?
I think you can change it to:
```
if (startNode == null) {
     return false;
}
return changeNodeToken(...) // since it's already a boolean.
```

src/main/java/com/google/sps/servlets/DataServlet.java
IshaniSanturkar: It's because the ones in the proto object are immutable so trying to modify them in add token and remove token was not allowed.

src/main/java/com/google/sps/servlets/DataServlet.java
IshaniSanturkar: I thought it would fall through and return false from the default because there's no break but to be consistent with the others I'll change it to that :)

EatingW: Ah I missed that there was no break :)

src/main/java/com/google/sps/servlets/DataServlet.java
EatingW: Thanks for explaining!

PR #5, opened 2020-07-01
Add unit tests for graph generation and mutation application

IshaniSanturkar:
In this PR, we:
- added unit tests to verify that the Guava graph was correctly being generated from a proto string-Node map
- added unit tests to verify that our code for processing various mutations was correct.

### DIFF
--- a/src/test/java/com/google/sps/MutationTest.java
+++ b/src/test/java/com/google/sps/MutationTest.java
@@ -15,21 +15,15 @@
 package com.google.sps;
 
 import com.google.common.graph.*;
-import com.google.protobuf.Struct;
-import com.google.protobuf.Value;
 import com.google.sps.servlets.DataServlet;
 import java.util.ArrayList;
 import java.util.Set;
 import java.util.HashMap;
-import java.util.Map;
 import java.util.List;
-import java.util.Iterator;
 import com.google.sps.data.GraphNode;
-import com.proto.GraphProtos.Graph;
 import com.proto.GraphProtos.Node;
 import com.proto.GraphProtos.Node.Builder;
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/sps/MutationTest.java
```diff
@@ -0,0 +1,539 @@
...
+            .setType(Mutation.Type.CHANGE_TOKEN)
+            .setTokenChange(tokenMut)
+            .build();
+
+    boolean success = servlet.mutateGraph(addTokenToA, graph, graphNodesMap);
+    Assert.assertTrue(success);
+
+    Set<GraphNode> graphNodes = graph.nodes();
+    Assert.assertTrue(graphNodesMap.containsKey("A"));
+    GraphNode newNodeA = graphNodesMap.get("A");
```
EatingW, line 383: Do you expect gnodeA to be reference equal to newNodeA? If so, also add an assert for this.

src/test/java/com/google/sps/MutationTest.java
```diff
@@ -0,0 +1,539 @@
...
+            .setType(Mutation.Type.CHANGE_TOKEN)
+            .setTokenChange(tokenMut)
+            .build();
+
+    boolean success = servlet.mutateGraph(removeTokenFromA, graph, graphNodesMap);
+    Assert.assertTrue(success);
+
+    Set<GraphNode> graphNodes = graph.nodes();
+    Assert.assertTrue(graphNodesMap.containsKey("A"));
+    GraphNode newNodeA = graphNodesMap.get("A");
```
EatingW, line 443: Same here, is newNodeA the same object as gnodeA?

PR #10, opened 2020-07-06
Add code to convert graph to JSON, send it to client and display it on the page

IshaniSanturkar:
In this PR, we:

Added code to convert the graph's nodes and edges into a JSON representation
Transmit the JSON to the client
Check for errors and display appropriate error messages
Parse the received response and render it on the page as a cytoscape graph
Success:
Screenshot 2020-07-01 at 3 30 28 PM
Failure:
Screenshot 2020-07-01 at 3 29 18 PM


EatingW:
BTW remember to review each others code first! @connorbrett @haoyxc

src/main/webapp/script.js
```diff
@@ -12,7 +12,118 @@
...
+  const serverErrorStatus = response.headers.get("serverError");
+
+  // Error on server side
+  if (serverErrorStatus !== null) {
+    displayError(serverErrorStatus);
+  }
+
+  const jsonResponse = await response.json();
+  // Graph nodes and edges received from server
+  let nodes = jsonResponse[0];
```
EatingW, line 35: Instead of hardcoding index [0] and [1] here, it seems less error prone to specify the Json data as {"nodes": nodeJson, "edges": edgeJson}

src/main/webapp/script.js
```diff
@@ -12,7 +12,118 @@
...
 
-function testFunc() {
-  fetch("/data").then(response => true);
-}
\ No newline at end of file
+  const serverErrorStatus = response.headers.get("serverError");
+
+  // Error on server side
+  if (serverErrorStatus !== null) {
+    displayError(serverErrorStatus);
```
EatingW, line 30: This displays the error but does not terminate function execution. Missing a return statement?

src/main/webapp/script.js
```diff
@@ -12,7 +12,118 @@
...
+        group: "edges",
+        data: {
+          id: `edge${start}${end}`,
+          target: end,
+          source: start
+        }
+      });
+    })
+    getGraphDisplay(graphNodes, graphEdges);
+    return;
```
EatingW, line 59: Instead of 
```
if (valid) {
  // do valid stuff
  return;
}
displayError
```
It is more straightforward if you do 
```
if (invalid) {
   displayError
   return;
}
// do valid stuff
```

src/main/webapp/index.html
```diff
@@ -5,13 +5,15 @@
...
 </head>
 
-<body id="content" onload="testFunc()">
+<body id="content">
   <h1>
-    HELLO
+    Graph Visualization Tool
   </h1>
+  <div id="graph"></div>
+  <script src="script.js"></script>
```
EatingW, line 16: Instead of embedding this script in the HTML, it's a better separation of concerns if you place all JS in the <head> section and call `generateGraph()` on body load.

src/test/java/com/google/sps/GraphGenerationTest.java
```diff
@@ -0,0 +1,195 @@
...
+import com.proto.GraphProtos.Node.Builder;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** */
+@RunWith(JUnit4.class)
+public final class GraphGenerationTest {
```
EatingW, line 36: Looks like this file and MutationTest is already merged into master? Likely this branch needs to pull from master.

src/main/webapp/index.html
```diff
@@ -5,13 +5,15 @@
   <meta charset="UTF-8">
   <title>Graphs</title>
   <link rel="stylesheet" href="style.css">
-  <script src="script.js"></script>
+  <script src="cytoscape/cytoscape.min.js"></script>
```
EatingW, line 8: Since you are only using `min.js` do you need to check in the other versions `.umd.js`, `.esm.js`, `.cjs.js`?

fenghaolw: Instead of checking this in, we can directly include it from one of the CDNs.

See https://js.cytoscape.org/#getting-started/including-cytoscape.js

src/main/webapp/script.js
haoyxc: So I'm having a bit of trouble figuring out the Json string. 

The first time I forgot the \" around nodes and the second time I forgot the \" around the nodeJson/edgeJson string. Currently I have: `String bothJson = "{\"nodes\": " + "\"" + nodeJson + "\""  + "," + "\"edges\": " + "\"" + edgeJson + "\"" + "}";` (which I can clean up a bit)

The toJson on the client works if I do something like
` String nodeJson = "hi";
 String edgeJson = "bye";`
but it doesn't with the actual nodeJson and edgeJson strings. Do you know why this could be the case?

src/main/webapp/index.html
IshaniSanturkar: In a different branch I also added package management using npm since a lot of the cytoscape plugins are included that way. Do you think importing cytoscape that way would be okay?

src/main/webapp/script.js
haoyxc: Just kidding, I used the JSONObject library to construct the object instead of doing it manually!

haoyxc: do you think we still need to do tests for JSON if we use a library?

src/main/webapp/index.html
haoyxc: OH i should've said this before resolving, but I used https://cdnjs.cloudflare.com/ajax/libs/cytoscape/3.15.1/cytoscape.min.js since it said it was the default. Would this be compatible with Ishani's?

src/main/webapp/script.js
EatingW: Having a small test doesn't hurt :) It doesn't have to be in this PR though.

src/main/webapp/index.html
EatingW: Cindy's link should be the same file as the previously check-in version so I think this is fine!

EatingW:
@haoyxc Cindy I think somehow googlebot doesn't even recognize your CLA? Did you switch an email for the commits? For example googlebot was happy with commit 1eeb6ff but didn't accept your most recent commits.

EatingW:
> @haoyxc Cindy I think somehow googlebot doesn't even recognize your CLA? Did you switch an email for the commits? For example googlebot was happy with commit [1eeb6ff](https://github.com/googleinterns/step49-2020/commit/1eeb6ff2f8d967ca4a1ac7d5a2dbc16d1c5eee3a) but didn't accept your most recent commits.

Looks like for some commits you used cindyup@gmail.com. Could you sign the CLA using that email as well?

haoyxc:
> > @haoyxc Cindy I think somehow googlebot doesn't even recognize your CLA? Did you switch an email for the commits? For example googlebot was happy with commit [1eeb6ff](https://github.com/googleinterns/step49-2020/commit/1eeb6ff2f8d967ca4a1ac7d5a2dbc16d1c5eee3a) but didn't accept your most recent commits.
> 
> Looks like for some commits you used [cindyup@gmail.com](mailto:cindyup@gmail.com). Could you sign the CLA using that email as well?

image
Is it this?

EatingW:
@googlebot Please rescan :)

haoyxc:
@googlebot I fixed it.

PR #20, opened 2020-07-08
refactor DataServlet to avoid task duplication

haoyxc:
The following tasks are performed just once: 

- Reading the graph file
- Reading the mutations file
- Applying all the mutations
- Converting the proto graph to a guava graph

Also add an immutable graph that's set to the original graph, prior to any mutations. (could be useful for displaying mutations later on)


src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -42,6 +42,31 @@
...
+  private MutableGraph<GraphNode> graph = null;
+  private HashMap<String, GraphNode> graphNodesMap = null;
+
+  private ImmutableGraph<GraphNode> graphOriginal = null; // never undergoes mutations
+
+  // Data structure that stores the roots of the graph across mutations
+  // Roots are nodes with no in-edges
+  private HashSet<String> roots = null;
+
+  private List<Mutation> mutList = null;
```
fenghaolw, line 68: I feel like this could be pulled into a separate data class? :)

haoyxc: Do you mean a data class with roots and a list of mutations as its fields?

fenghaolw: I mean all the protoGraph, protoNodesMap, graph, graphNodesMap, graphOriginal, roots, mutList could be in a separate data class (or maybe two classes, one for graph and one for mutations).

This decouples the "data class" and the servlet, even if we still need to construct the data class within the get method, it is easy to change that in the future. Does this make sense?

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -42,6 +42,31 @@
 
 @WebServlet("/data")
 public class DataServlet extends HttpServlet {
+  // PROTO Data structure:
+  // Parse the contents of graph.txt into a proto Graph object, and extract
+  // information
+  // from the proto object into a map. This is used to store the proto input and
+  // isn't updated
+  // with mutations.
+  private Graph protoGraph = null;
```
fenghaolw, line 51: Do we need to store protoGraph and protoNodesMap? Seems like these two are only used when we parse the graph. All the other places only consume graphOriginal and graph.

haoyxc: yea you're right, we don't. thanks for pointing it out!

src/main/java/com/google/sps/servlets/DataServlet.java
haoyxc: yea that makes sense!

src/main/java/com/google/sps/data/DataGraph.java
```diff
@@ -0,0 +1,125 @@
+package com.google.sps.data;
+
+// import java.util.ArrayList;
```
fenghaolw, line 3: Revert these comments?

src/main/java/com/google/sps/data/DataGraph.java
```diff
@@ -0,0 +1,125 @@
...
+import com.proto.GraphProtos.Graph;
+import com.proto.GraphProtos.Node;
+import java.util.HashMap;
+import java.util.List;
+import java.util.ArrayList;
+import com.google.protobuf.Struct;
+import java.util.Map;
+import java.util.HashSet;
+
+public class DataGraph {
```
fenghaolw, line 29: Why not use AutoValue? :)

src/main/java/com/google/sps/data/DataGraph.java
```diff
@@ -0,0 +1,125 @@
...
+        // This child can no longer be a root since it has an in-edge
+        roots.remove(child);
+        GraphNode childNode = protoNodeToGraphNode(protoNodesMap.get(child));
+        if (!graphNodesMap.containsKey(child)) {
+          // If child node is not already in the graph, add it
+          graph.addNode(childNode);
+          graphNodesMap.put(child, childNode);
+        } else if (graph.hasEdgeConnecting(childNode, graphNode)) {
+          // the graph is not a DAG, so we error out
+          return false;
```
fenghaolw, line 103: You don't use the return value here. Should be fine to throw an IllegalArgumentException

src/main/java/com/google/sps/data/Utility.java
```diff
@@ -0,0 +1,234 @@
...
+import java.util.ArrayList;
+import com.google.protobuf.Struct;
+import java.util.Map;
+import java.util.HashSet;
+import java.util.Set;
+import com.google.common.reflect.TypeToken;
+import java.lang.reflect.Type;
+import org.json.JSONObject;
+
+public class Utility {
```
fenghaolw, line 20: Make this final to avoid inheritance.

Also add a private constructor to enforce noninstantiability, since this class is supposed to be a "utility class" that carries a bunch of static methods.

src/main/java/com/google/sps/data/Utility.java
```diff
@@ -0,0 +1,234 @@
...
+            .toString();
+    return allJson;
+  }
+
+  /*
+   * Changes the graph according to the given mutation object. The parameters are
+   * mutated in place.
+   *
+   * @param mut the mutation to affect
+   *
```
fenghaolw, line 50: nit: usually there is no newlines between @param

src/main/java/com/google/sps/data/DataGraph.java
IshaniSanturkar: @fenghaolw even though I pushed some code to this branch right now, the code hasn't been modified to address your comments fully. We'll comment here once it's ready for your review :)

src/main/java/com/google/sps/data/DataGraph.java
haoyxc: image
We use it in the DataServlet I believe. Do you think an IllegalArgumentException would be a better way of handling it?

haoyxc: (it's the value of success!)

src/main/java/com/google/sps/data/DataGraph.java
haoyxc: My understanding may be incorrect, but I thought the parameters passed into the constructor were also the private instance variables stored in AutoValue! And that's not what we want, right?

src/main/java/com/google/sps/data/DataGraph.java
haoyxc: Other than the replies to your comments for what I wasn't sure about, I think I addressed them! :)

src/main/java/com/google/sps/data/DataGraph.java
fenghaolw: It cannot be fully private, but could be package-visible only. I think that is sufficient for encapsulation.

https://github.com/google/auto/blob/master/value/userguide/howto.md#-keep-my-accessor-methods-private

The pros is to avoid boilerplates (equals/hashcode/getter/setter) and mistakes. IDE can certainly generate and maintain these boilerplates but then it is still visible to readers.

src/main/java/com/google/sps/data/DataGraph.java
fenghaolw: Sorry I miss that. I think setting the error response is better.

src/test/java/com/google/sps/MaxDepthTest.java
```diff
@@ -71,21 +70,19 @@ public void maxDepthZero() {
...
-    servlet.graphFromProtoNodes(protoNodesMap, graph, graphNodesMap, roots);
+    DataGraph dataGraph = new DataGraph();
+    dataGraph.graphFromProtoNodes(protoNodesMap);
+    MutableGraph<GraphNode> graph = dataGraph.getGraph();
+    HashMap<String, GraphNode> graphNodesMap = dataGraph.getGraphNodesMap();
+    HashSet<String> roots = dataGraph.getRoots();
 
     MutableGraph<GraphNode> truncatedGraph =
-        servlet.getGraphWithMaxDepth(graph, roots, graphNodesMap, 0);
+        Utility.getGraphWithMaxDepth(graph, roots, graphNodesMap, 0);
```
fenghaolw, line 85: usually it is better to import static method instead of import the entire class.

`import static com.google.sps.data.Utility.getGraphWithMaxDepth;`

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -57,396 +56,56 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-        // The number of outgoing edges from the current node, number of nodes in the
-        // next layer
-        for (GraphNode gn : graphInput.successors(curr)) {
-          if (!visited.containsKey(gn)) {
-            queue.add(gn);
-            nextDepthElementCount++;
-          }
+    MutableGraph<GraphNode> graph = currDataGraph.getGraph();
+    HashMap<String, GraphNode> graphNodesMap = currDataGraph.getGraphNodesMap();
+    HashSet<String> roots = currDataGraph.getRoots();
```
EatingW, line 87: getRoots() returns a copy of the roots? Then this means that if the mutation changes a root, it won't actually be applied to the graph itself, only to the copied `roots` list, so after applying the mutation, if you call `currDataGraph.getRoots()` the roots will be outdated. 

I think currDataGraph should always be reflecting the correct state of information (graphNodes, roots) so you should probably not create a copy on the getters. 

This can also be solved by moving the mutateGraph function into DataGraph, which you are doing in a followup PR, right?

IshaniSanturkar: Yup and we are also returning the roots directly with the autovalue methods.

src/main/java/com/google/sps/Utility.java
```diff
@@ -1,4 +1,4 @@
-package com.google.sps.data;
+package com.google.sps;
```
EatingW, line 1: Still missing license :)

haoyxc:
@EatingW do you think you could approve this again? Adding the license dismissed your approval -- so sorry!!

PR #22, opened 2020-07-09
Add next and previous buttons and corresponding front-end code and tests

IshaniSanturkar:
Updated: 
- added next and previous buttons to the HTML page which when clicked generate the next/prev graph in the sequence of mutations
- added javascript code to record clicks on these buttons and submit a request to the server to generate the new graph
- added tests to verify the correctness of the JS code.
--------------
- created a separate function to set the numMutations and currNumMutations
- tested those as well!

image


src/main/webapp/script.js
```diff
@@ -90,18 +98,25 @@ async function generateGraph() {
  * Ensures that the depth is an integer between 0 and 20
  */
 function getUrl() {
-  let selectedDepth = document.getElementById('num-layers').value;
-  if (selectedDepth.length === 0) {
+  const depthElem = document.getElementById('num-layers');
+  let selectedDepth = 0;
```
haoyxc, line 102: is there a reason you needed to define this variable?

src/main/webapp/script.js
```diff
@@ -236,3 +251,16 @@ function getTooltipContent(node) {
 
   return content;
 }
+
+/**
+ * When a next/previous button is clicked, modifies the mutation index of the
+ * current graph to represent the new state. Then, the corresponding
+ * graph is requested from the server.
+ */
+function navigateGraph(amount) {
```
haoyxc, line 260: what if the amount is greater than the number of mutations?

src/main/webapp/script.js
IshaniSanturkar: Just in case the num-layers input field isn't defined so that I can check if it's null before trying to get its value. This came up in one of my tests where I have to set up the document by hand and forgot to initialize this field so I was getting an error. Do you think I should leave the check or modify my tests to comply?

src/main/webapp/script.js
IshaniSanturkar: I added a parameter in another branch (mutBackend) for this because there it would have a meaningful value. Do you think I should add it here anyways?

src/main/webapp/script.js
haoyxc: oh it's not a big deal haha should be good the way it is!

src/main/webapp/script.js
haoyxc: I think it would be useful for the user to know when they're at the end! right now does the user know if they're at 0 or at the end? (maybe instead of the roots you could pass mutList.size() to the client?) and then display something when they're either at 0 or at end

IshaniSanturkar: That's a good point. So right now, I'm going to add a dummy value to the JS for the maximum next clicks they can make. In the next PR I will actually populate this value from the backend :)

haoyxc:
nit: currently the arrows don't fit on the screen (within the viewport). Maybe make the graph smaller?

src/main/webapp/script.js
```diff
@@ -236,3 +255,39 @@ function getTooltipContent(node) {
...
+  currGraphNum += amount;
+  if(currGraphNum < 0) {
```
haoyxc, line 266: return out of it so you don't have to regenerate the graph? (i did this in mutBackend so if you don't it's nbd). But adding something to tell the user they're at the end would be nice!

src/main/webapp/script.js
```diff
@@ -236,3 +255,39 @@ function getTooltipContent(node) {
...
+    currGraphNum = 0;
+  }
+  if(currGraphNum >= numMutations) {
+    currGraphNum = numMutations;
+  }
+  generateGraph();
```
haoyxc, line 272: same thing

haoyxc: just kidding my bad!!

spec/jstests-spec.js
```diff
@@ -195,3 +195,75 @@ describe("Initializing tooltips", function() {
...
+    prevButton.id = "prevbutton";
+    prevButton.onclick = () => navigateGraph(-1);
+    const nextButton = document.createElement("button");
+    nextButton.id = "nextbutton";
+    nextButton.onclick = () => navigateGraph(1);
+    document.body.appendChild(prevButton);
+    document.body.appendChild(nextButton);
+
+    expect(currGraphNum).toBe(0);
+    expect(numMutations).toBe(3);
```
EatingW, line 211: This currGraphNum and numMutations are both globally exported variables. If you end up changing the value of numMutations, you'd have to make sure to change this test as well. 
Instead, why not export `initializeNumMutations(num)` that sets this variable? You then have the flexibility to set it to anything in tests, and can also call this method when you actually pass in the total number of mutations to script.js? 
Same goes for initializing currGraphNum.

spec/jstests-spec.js
```diff
@@ -195,3 +195,75 @@ describe("Initializing tooltips", function() {
...
+
+    nextButton.click();
+    expect(currGraphNum).toBe(1);
+
+    nextButton.click();
+    expect(currGraphNum).toBe(2);
+
+    nextButton.click();
+    expect(currGraphNum).toBe(3);
```
EatingW, line 220: nextButton should already be disabled here? It might also make sense to assert that the buttons are not disabled in previous clicks.

spec/jstests-spec.js
```diff
@@ -195,3 +195,75 @@ describe("Initializing tooltips", function() {
...
+
+    nextButton.click();
+    expect(currGraphNum).toBe(3);
+
```
EatingW, line 224: nit: delete empty line

spec/jstests-spec.js
```diff
@@ -195,3 +195,75 @@ describe("Initializing tooltips", function() {
...
+    expect(prevButton.disabled).toBe(false);
+
+    nextButton.click();
+    expect(currGraphNum).toBe(2);
+
+    prevButton.click();
+    expect(currGraphNum).toBe(1);
+
+    prevButton.click();
+    expect(currGraphNum).toBe(0);
```
EatingW, line 245: Similarly, assert that the prevButton is disabled starting at this point

spec/jstests-spec.js
```diff
@@ -195,3 +195,75 @@ describe("Initializing tooltips", function() {
...
+
+    prevButton.click();
```
EatingW, line 247: I think you can delete this test - same functionality as lines 250-253

spec/jstests-spec.js
```diff
@@ -195,3 +195,75 @@ describe("Initializing tooltips", function() {
...
+    expect(currGraphNum).toBe(0);
+    expect(nextButton.disabled).toBe(false);
+    expect(prevButton.disabled).toBe(true);
+ 
+    nextButton.click();
+    expect(currGraphNum).toBe(1);
+    expect(nextButton.disabled).toBe(false);
+    expect(prevButton.disabled).toBe(false);
+
+    const requestString = getUrl();
```
EatingW, line 260: Asserting that the correct mutationNum is passed should be its own test.

src/main/webapp/script.js
```diff
@@ -236,3 +255,39 @@ function getTooltipContent(node) {
...
+  updateButtons();
```
EatingW, line 273: One thing to consider here is what happens when the graph takes a long time to load, and the user keeps pressing next/prev impatiently? There would be multiple calls to navigateGraph then. 

Maybe we should prevent this from happening by disabling both buttons when generate graph is happening? Maybe also have an indicator (e.g. spinning wheel)? 

Does not have to be addressed by this PR!

spec/jstests-spec.js
haoyxc: moved everything under that line into a new test!

spec/jstests-spec.js
haoyxc: I added tests for the next/prev buttons after every click, hope it wasn't too repetitive! :)

spec/jstests-spec.js
haoyxc: Ohh that's smart, and I think I understand what you mean! Made the changes, but let me know if it's not what you were expecting!

src/main/webapp/script.js
haoyxc: I didn't know where else this could go, so I disabled both before the fetch request and enabled both before the return statement in generateGraph. Do you think there's a better way to do it, or would this be okay?

spec/jstests-spec.js
```diff
@@ -255,13 +258,25 @@ describe("Pressing next and previous buttons associated with a graph", function(
...
     expect(nextButton.disabled).toBe(false);
-    expect(prevButton.disabled).toBe(false);
+    expect(prevButton.disabled).toBe(false);  
+  });
+});
 
-    const requestString = getUrl();
-    const requestParams = requestString.substring(requestString.indexOf("?"));
+describe("Check initializing variables are passed correctly", function() {
+    beforeEach(function () {
```
EatingW, line 266: nit: indentation

spec/jstests-spec.js
```diff
@@ -255,13 +258,25 @@ describe("Pressing next and previous buttons associated with a graph", function(
...
+    setCurrGraphNum(1);
+  });
+
+  afterEach(function () {
+    setCurrGraphNum(0);
+  });
+  it("passes correct value of the mutations number in the fetch request", function() {
```
EatingW, line 273: nit: newline before `it` and no newline after.

spec/jstests-spec.js
```diff
@@ -255,13 +258,25 @@ describe("Pressing next and previous buttons associated with a graph", function(
...
+
+    let requestString = getUrl();
```
EatingW, line 275: Why `let`, not `const`?

spec/jstests-spec.js
```diff
@@ -255,13 +258,25 @@ describe("Pressing next and previous buttons associated with a graph", function(
...
+    let requestParams = requestString.substring(requestString.indexOf("?"));
     
-    const constructedUrl = new URLSearchParams(requestParams);
-    expect(constructedUrl.has("depth")).toBe(true);
+    let constructedUrl = new URLSearchParams(requestParams);
```
EatingW, line 278: Why `let`, not `const`?

spec/jstests-spec.js
```diff
@@ -255,13 +258,25 @@ describe("Pressing next and previous buttons associated with a graph", function(
...
+      expect(constructedUrl.has("depth")).toBe(true);
```
EatingW, line 279: nit: indentation

src/main/webapp/script.js
```diff
@@ -48,7 +62,15 @@ async function generateGraph() {
   let graphNodes = [];
   let graphEdges = [];
 
+  // disable buttons
+  const prevBtn = document.getElementById("prevbutton");
+  const nextBtn = document.getElementById("nextbutton");
+
   const url = getUrl();
+
+  prevBtn.disabled = true;
```
EatingW, line 71: Move these two lines to before `getUrl`?

src/main/webapp/script.js
```diff
@@ -264,10 +287,12 @@ function getTooltipContent(node) {
 function navigateGraph(amount) {
   currGraphNum += amount;
   if(currGraphNum < 0) {
-    currGraphNum = 0;
+    setCurrGraphNum(0);
```
EatingW, line 290: I don't think you need to call this function - you can just set currGraphNum directly. 
Alternatively, you can make the setCurrGraphNum function clamp the values to 0 and numMutations, and just call `setCurrGraphNum(currGraphNum + amount)`

src/main/webapp/script.js
EatingW: I think this is fine! You might want to play with it locally a bit to see if this looks good to you from a user experience perspective. For example, you can manually introduce a 10 second delay to the generateGraph function and see if interaction with buttons make sense. 

Since generateGraph already calls updateButtons, delete this line?

src/main/webapp/script.js
```diff
@@ -264,10 +287,12 @@ function getTooltipContent(node) {
...
+    return;
   }
-  if(currGraphNum >= numMutations) {
-    currGraphNum = numMutations;
+  if(currGraphNum > numMutations) {
```
EatingW, line 293: nit: space after `if`

spec/jstests-spec.js
IshaniSanturkar: Whoops, sorry about that

src/main/webapp/script.js
IshaniSanturkar: I found that calling generateGraph from the test file was a bit problematic because it is asynchronous but our tests run before it finishes so the buttons might just be globally disabled causing the test to fail. I changed it so that in the HTML, I call navigate graph and then generate graph (which calls update buttons) and in the tests, I call navigate graph and update buttons on button click. Does that sound okay?

EatingW:
Remember to merge PR21 before this one and change the PR to merge to master instead! Otherwise PR21 would get too large.

haoyxc:
> Remember to merge PR21 before this one and change the PR to merge to master instead! Otherwise PR21 would get too large.

Done!

PR #23, opened 2020-07-09
Searchnode

connorbrett:
search for node by node ID and zoom in if it exists


haoyxc:
pull from master or rebase onto refactorBackend (I think is the name!)

src/main/webapp/script.js
```diff
@@ -173,6 +173,42 @@ function getGraphDisplay(graphNodes, graphEdges) {
     const node = evt.target;
     node.tip.show();
   });
+
+  let searchElement = document.getElementById('search');
```
haoyxc, line 177: maybe use const because searchElement shouldn't change?

src/main/webapp/script.js
```diff
@@ -173,6 +173,42 @@ function getGraphDisplay(graphNodes, graphEdges) {
...
+  searchElement.onchange = function() {
+    if (searchNode(cy, searchElement.value)) {
+      document.getElementById('searcherror').innerText = "";
```
haoyxc, line 180: if searcherror is empty by default, why is this necessary?

src/main/webapp/script.js
```diff
@@ -173,6 +173,42 @@ function getGraphDisplay(graphNodes, graphEdges) {
...
+    } else {
+      document.getElementById('searcherror').innerText = "Node does not exist.";
```
haoyxc, line 182: What happens if the user deletes what's in the search? (then the name is empty so the node isn't found, right?) But we don't want the "Node does not exist"?

src/main/webapp/script.js
```diff
@@ -173,6 +173,42 @@ function getGraphDisplay(graphNodes, graphEdges) {
     const node = evt.target;
     node.tip.show();
   });
+
+  let searchElement = document.getElementById('search');
+  searchElement.onchange = function() {
```
haoyxc, line 178: You should consider add tests for this! To make sure the innerText is being updated correctly?

src/main/webapp/script.js
```diff
@@ -173,6 +173,42 @@ function getGraphDisplay(graphNodes, graphEdges) {
...
+    }
+  };
+  
+}
+
+/**
+ * Zooms in on specific node
+ */
+function searchNode(cy, query) {
+  let target = findNodeInGraph(cy, query);
```
haoyxc, line 192: target shouldn't change, so why not consider using const?

src/main/webapp/script.js
```diff
@@ -173,6 +173,42 @@ function getGraphDisplay(graphNodes, graphEdges) {
...
+  if (target) {
+    cy.fit(target, 50);
```
haoyxc, line 194: Could you also blur the surrounding nodes and maybe highlight which node is the target node? (because when I tried running it, there were a lot of nodes on the screen and it was kind of difficult to see!)

src/main/webapp/index.html
```diff
@@ -19,6 +19,9 @@ <h1>
   <button onclick="graph.generateGraph()">Get Graph</button>
   <br/>
   <div id="graph"></div>
+  <label>Search by Node: </label>
+  <input type="text" id="search" name="search"></input>
+  <label id="searcherror"></label>
```
haoyxc, line 24: https://google.github.io/styleguide/htmlcssguide.html#ID_and_Class_Name_Delimiters
maybe consider using search-error?

src/main/webapp/script.js
connorbrett: I have this so if someone enters an invalid node, then a valid one, the error message will go away.

haoyxc: maybe you could consider adding a button rather than doing it onChange. You probably don't need to update the graph every time the box changes (since typing out a node name could take a while?)

src/main/webapp/script.js
connorbrett: If there is nothing in the search, it will essentially fit all of the nodes on screen (so just reset) so the error message won't show.

src/main/webapp/script.js
connorbrett: couldn't figure out how to do that with cytoscape but I made it so it changed the color of the node so its more obvious now

haoyxc: maybe look at the wine ex that ishani brought up?

src/main/webapp/script.js
```diff
@@ -173,6 +173,46 @@ function getGraphDisplay(graphNodes, graphEdges) {
     const node = evt.target;
     node.tip.show();
   });
+
+  const searchElement = document.getElementById('search');
+  searchElement.onchange = function() {
```
EatingW, line 178: Why not use a button for this? If the graph was very large, performing the operation could cause a noticeable delay.

src/main/webapp/script.js
```diff
@@ -173,6 +173,46 @@ function getGraphDisplay(graphNodes, graphEdges) {
...
+
+/**
+ * Finds element in cy graph by id
+ */
+function findNodeInGraph(cy, id) {
+  const target = cy.$('#'+id);
+  if (target.length != 0 && id.length != 0) {
+    return target;
+  } else {
+    return null;
```
EatingW, line 214: nit: move the return statement outside of the if-block.

PR #24, opened 2020-07-10
Refactor backend to make functions take in a data graph instead of taking in its individual fields

IshaniSanturkar:
In this PR, we:
- moved some utility functions to be instance functions for a data graph
- restricted access to some of the graph's fields by making a copy of them before returning them to the user
- modified the tests and code to comply with the new structure.

### DIFF
--- a/src/main/java/com/google/sps/MutationTest.java
+++ b/src/main/java/com/google/sps/MutationTest.java
@@ -74,7 +74,6 @@ public final class MutationTest {
     Mutation addA = Mutation.newBuilder().setType(Mutation.Type.ADD_NODE).setStartNode("A").build();
 
     HashSet<String> roots = new HashSet<>();
-
     DataGraph dataGraph = DataGraph.create(graph, graphNodesMap, roots);
 
     boolean success = dataGraph.mutateGraph(addA);
@@ -98,12 +97,13 @@ public final class MutationTest {
     graphNodesMap.put("A", gNodeA);
 
     Mutation addA = Mutation.newBuilder().setType(Mutation.Type.ADD_NODE).setStartNode("A").build();
+
     HashSet<String> roots = new HashSet<>();
     DataGraph dataGraph = DataGraph.create(graph, graphNodesMap, roots);
 
     boolean success = dataGraph.mutateGraph(addA);
-
     Assert.assertTrue(success);
+
     Set<GraphNode> graphNodes = graph.nodes();
     Assert.assertEquals(graphNodes.size(), 1);
     Assert.assertTrue(graphNodes.contains(gNodeA));
@@ -128,8 +128,10 @@ public final class MutationTest {
             .setStartNode("A")
             .setEndNode("B")
             .build();
+
     HashSet<String> roots = new HashSet<>();
     DataGraph dataGraph = DataGraph.create(graph, graphNodesMap, roots);
+
     boolean success = dataGraph.mutateGraph(addAB);
     Assert.assertTrue(success);
 
@@ -139,7 +141,6 @@ public final class MutationTest {
     Assert.assertTrue(graphNodes.contains(gNodeB));
     Assert.assertEquals(graphNodesMap.get("A"), gNodeA);
     Assert.assertEquals(graphNodesMap.get("B"), gNodeB);
-
     Assert.assertTrue(graph.hasEdgeConnecting(gNodeA, gNodeB));
     Assert.assertFalse(graph.hasEdgeConnecting(gNodeB, gNodeA));
   }
@@ -167,8 +168,8 @@ public final class MutationTest {
 
     HashSet<String> roots = new HashSet<>();
     DataGraph dataGraph = DataGraph.create(graph, graphNodesMap, roots);
-    boolean success = dataGraph.mutateGraph(addAB);
 
+    boolean success = dataGraph.mutateGraph(addAB);
     Assert.assertFalse(success);
 
     Set<GraphNode> graphNodes = graph.nodes();
@@ -177,7 +178,6 @@ public final class MutationTest {
     Assert.assertTrue(graphNodes.contains(gNodeB));
     Assert.assertEquals(graphNodesMap.get("A"), gNodeA);
     Assert.assertEquals(graphNodesMap.get("B"), gNodeB);
-
     Assert.assertTrue(graph.hasEdgeConnecting(gNodeA, gNodeB));
   }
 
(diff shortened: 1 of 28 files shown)


src/main/java/com/google/sps/data/Utility.java
```diff
@@ -61,179 +59,39 @@ public static String graphToJson(MutableGraph<GraphNode> graph, HashSet<String>
...
-    }
-    for (EndpointPair<GraphNode> edge : graphInput.edges()) {
-      if (graphToReturn.nodes().contains(edge.nodeU())
-          && graphToReturn.nodes().contains(edge.nodeV())) {
-        graphToReturn.putEdge(edge.nodeU(), edge.nodeV());
-      }
-    }
-
-    return graphToReturn;
+  public static HashSet<String> copyRoots(HashSet<String> roots) {
```
haoyxc, line 78: Why is this in utility? I think it makes more sense to have a generic copySet function here? Bc if you just copy the roots that seems specific to the DataGraph and should probably go there

src/main/java/com/google/sps/data/Utility.java
```diff
@@ -61,179 +59,39 @@ public static String graphToJson(MutableGraph<GraphNode> graph, HashSet<String>
...
-    if (depthRemaining >= 0) {
-      visited.put(gn, true);
-      graphToReturn.addNode(gn);
-      for (GraphNode child : graphInput.successors(gn)) {
-        if (!visited.containsKey(child)) {
-          // Visit the child and indicate the increase in depth
-          dfsVisit(child, graphInput, visited, graphToReturn, depthRemaining - 1);
-        }
-      }
+  public static HashMap<String, GraphNode> copyNodeMap(HashMap<String, GraphNode> graphNodesMap) {
```
haoyxc, line 90: Same thing as roots, maybe a generic copyHashMap function

src/main/java/com/google/sps/data/Utility.java
```diff
@@ -38,11 +34,13 @@ public static GraphNode protoNodeToGraphNode(Node thisNode) {
...
-   * contains nodes, edges, and the roots of the graph.
+  /**
+   * Converts a Guava graph into a String encoding of a JSON Object. The object contains nodes,
+   * edges, and the roots of the graph.
    *
    * @param graph the graph to convert into a JSON String
+   * @param roots the roots of the graph to convert into a JSON String
+   * @return a JSON object containing as entries the nodes, edges and roots of this graph
    */
   public static String graphToJson(MutableGraph<GraphNode> graph, HashSet<String> roots) {
```
haoyxc, line 45: Put this in DataGraph as well? Because you can just get the graph and the roots from the DataGraph object

src/main/java/com/google/sps/data/Utility.java
```diff
@@ -61,179 +59,39 @@ public static String graphToJson(MutableGraph<GraphNode> graph, HashSet<String>
...
-    if (tokenMutType == TokenMutation.Type.ADD_TOKEN) {
-      tokenList.addAll(tokenNames);
-    } else if (tokenMutType == TokenMutation.Type.DELETE_TOKEN) {
-      tokenList.removeAll(tokenNames);
-    } else {
-      // unrecognized mutation
-      return false;
-    }
-    return true;
+  public static MutableGraph<GraphNode> copyGraph(MutableGraph graph) {
```
haoyxc, line 68: Is there really a point of this function? It seems straightforward enough to say Graphs.copyOf(graph) in DataGraph. I really don't think this is needed?

src/main/java/com/google/sps/data/GraphNode.java
```diff
@@ -28,4 +29,12 @@ public static GraphNode create(String name, List<String> tokenList, Struct metad
   public int hashCode() {
     return name().hashCode();
   }
+
+  // Returns a deep copy of the given graph node
+  public GraphNode getCopy() {
```
haoyxc, line 34: Test this? :) (to make sure it's not referentially equal)

src/main/java/com/google/sps/data/DataGraph.java
```diff
@@ -1,34 +1,57 @@
...
+    this.graphNodesMap = map;
+    this.roots = roots;
+  }
+
+  /**
+   * Returns a deep copy of the given data graph
+   *
+   * @return a deep copy of this data graph
+   */
+  public DataGraph getCopy() {
```
haoyxc, line 44: Also test this? to make sure it's not referentially equal

src/main/java/com/google/sps/data/Utility.java
haoyxc: and maybe add tests for this as well

src/main/java/com/google/sps/data/Utility.java
haoyxc: and same thing, add tests?

src/main/java/com/google/sps/data/DataGraph.java
```diff
@@ -1,34 +1,57 @@
...
+  }
+
   /**
    * Getter for the graph
    *
    * @return the graph
    */
   public MutableGraph<GraphNode> getGraph() {
-    return this.graph;
+    return Utility.copyGraph(this.graph);
```
haoyxc, line 54: this could similarly be accomplished with Graphs.copyGraph(this.graph) so I really don't see the point of calling another function?

haoyxc:
also since I did contribute to this & was familiar with much of it already @connorbrett feel free to also look at it so we have fresh eyes :)))

src/main/java/com/google/sps/data/Utility.java
IshaniSanturkar: I moved it back to DataServlet. Also, since I'm just using an API functions to copy the roots do you think I would still need to test?

src/main/java/com/google/sps/data/Utility.java
IshaniSanturkar: Moved it back to DataGraph, I'll add tests to make sure it's a deep copy.

src/main/java/com/google/sps/data/Utility.java
IshaniSanturkar: Yep removed it!

src/main/java/com/google/sps/MutationTest.java
```diff
@@ -74,7 +74,10 @@ public void addNodeBasic() {
     Mutation addA = Mutation.newBuilder().setType(Mutation.Type.ADD_NODE).setStartNode("A").build();
 
     HashSet<String> roots = new HashSet<>();
-    boolean success = Utility.mutateGraph(addA, graph, graphNodesMap, roots);
+
```
EatingW, line 77: nit: make these newlines consistent with other tests? The same applies to the other tests as well. For example, roots and dataGraph could always be a single block while mutateGraph and assert can be another block.

haoyxc:
Sorry @EatingW just changed the base to master!!

PR #26, opened 2020-07-13
Implement backend for displaying mutations

haoyxc:
Add a function to the backend that takes in a given number and generates a graph with that many mutations applied (so that the frontend arrows are functional!)

changes include:

- start graph at beginning (0 mutations applied)
- store the mutation number for a dataGraph 
- generalize getting mutations so that you can request any given mutation number, given the original graph and the current graph
- add tests for the mutation function

In action!
[capture (1).zip](https://github.com/googleinterns/step49-2020/files/4914761/capture.1.zip)


src/main/java/com/google/sps/Utility.java
```diff
@@ -51,15 +52,59 @@ public static GraphNode protoNodeToGraphNode(Node thisNode) {
...
+    } else {
+      // Create a copy of the original graph and start from the original graph
+      DataGraph originalCopy = original.getCopy();
+      for (int i = 0; i < mutationNum; i++) {
+        success = originalCopy.mutateGraph(mutList.get(i));
+        if (!success) {
+          return null;
+        }
+      }
+      return DataGraph.create(
```
EatingW, line 106: Same here, since originalCopy is already a copy, why not just return it?

src/main/java/com/google/sps/Utility.java
```diff
@@ -51,15 +52,59 @@ public static GraphNode protoNodeToGraphNode(Node thisNode) {
...
+            .toString();
     return resultJson;
   }
+
+  /**
+   * @param original the original graph
+   * @param curr the current (most recently-requested) graph
+   * @param mutationNum number of mutations to apply
+   * @param mutList mutation list
+   * @return the resulting data graph or null if there was an error Requires that original != curr
```
EatingW, line 79: The comment "Requires that original != curr" seems to belong somewhere else. Also consider asserting the precondition is met by using the Preconditions class https://guava.dev/releases/19.0/api/docs/com/google/common/base/Preconditions.html

src/main/java/com/google/sps/DataGraph.java
```diff
@@ -191,36 +225,59 @@ public boolean mutateGraph(Mutation mut) {
...
       tokenList.addAll(tokenNames);
     } else if (tokenMutType == TokenMutation.Type.DELETE_TOKEN) {
       tokenList.removeAll(tokenNames);
     } else {
       // unrecognized mutation
-      return false;
+      return null;
     }
-    return true;
+    GraphNode newNode = GraphNode.create(node.name(), tokenList, node.metadata());
```
EatingW, line 279: nit: merge these two lines to return GraphNode.create directly

src/main/java/com/google/sps/DataGraph.java
```diff
@@ -167,7 +201,7 @@ public boolean mutateGraph(Mutation mut) {
         }
         // Check whether any successor will have no in-edges after this node is removed
         // If so, make them roots
-        Set<GraphNode> successors = graph.successors(startNode);
+        successors = graph.successors(startNode);
```
EatingW, line 204: nit: you can just call graph.successors in the for loop header

src/main/java/com/google/sps/DataGraph.java
```diff
@@ -140,6 +172,8 @@ public boolean mutateGraph(Mutation mut) {
     GraphNode startNode = graphNodesMap.get(startName);
     GraphNode endNode = graphNodesMap.get(endName);
 
+    Set<GraphNode> successors;
```
EatingW, line 175: With the change I suggested below to place the `graph.successors` within the `if` header, you shouldn't need to declare the `successors` variable outside of the switch statement anymore. 

I assume you did it because otherwise it would be a compile error that variable `successors` is already defined. This is because Java case statements have a weird notion of scope - by default the `case` statements share a scope, which is a bit unintuitive. This is why I like to manually introduce a new scope for my case statements by adding extra `{}` so that you can do something like:
```
switch (mut.getType()) {
    case ADD_NODE: {
         int x = 1;
         break
    }
    case ADD_EDGE: {
         int x = 2;
         break
    }
}
```
The two `x`s above are now in different scopes :)

src/main/java/com/google/sps/DataGraph.java
```diff
@@ -191,36 +225,59 @@ public boolean mutateGraph(Mutation mut) {
...
+        graph.removeNode(startNode);
+
+        graph.addNode(newNode);
+        for (GraphNode succ : successors) {
+          graph.putEdge(newNode, succ);
+        }
+        for (GraphNode pred : predecessors) {
+          graph.putEdge(pred, newNode);
+        }
+
```
EatingW, line 247: nit delete extra newline

src/main/java/com/google/sps/GraphMutationTest.java
```diff
@@ -0,0 +1,300 @@
...
+import com.proto.MutationProtos.Mutation;
+import com.proto.MutationProtos.TokenMutation;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class GraphMutationTest {
```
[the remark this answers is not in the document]
IshaniSanturkar: Good idea!

src/main/java/com/google/sps/Utility.java
```diff
@@ -51,15 +52,59 @@ public static GraphNode protoNodeToGraphNode(Node thisNode) {
...
+
+    if (curr.numMutations() <= mutationNum) { // going forward
+      for (int i = curr.numMutations(); i < mutationNum; i++) {
+        // Mutate graph operates in place
+        success = curr.mutateGraph(mutList.get(i));
+        if (!success) {
+          return null;
+        }
+      }
+      return DataGraph.create(curr.graph(), curr.graphNodesMap(), curr.roots(), mutationNum);
```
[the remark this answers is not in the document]
IshaniSanturkar: The reason we did that was to modify the mutation number of the graph. Autovalue objects are immutable so we couldn't increment the mutation number variable, which is why we created a new object with identical fields except mutationNum. Do you think it would be better to store this integer outside the autovalue object?

src/main/java/com/google/sps/DataGraph.java
IshaniSanturkar: We were doing that originally. The issue is that before adding edges to the new node we have to delete the old node (and all edges to/from it) so when we get to the loop we can't access the successors and predecessors of the old node anymore.

src/main/java/com/google/sps/DataGraph.java
IshaniSanturkar: Oh that's a really good idea! Yeah I was so confused by that compile error, it's very unintuitive.

haoyxc: Oh whoa I did not know this!!

src/main/java/com/google/sps/DataGraph.java
haoyxc: By loop header do you mean before the switch statement?

haoyxc: I just tested it, and f i'm understanding this correctly, it's because the successors are only relevant in certain cases. So if the mutation is in adding a node and you call graph.successors(startNode) (and start not has yet to exist) you run into weird issues. (please correct me if my understanding of loop header is completely wrong!!)

src/main/java/com/google/sps/DataGraph.java
```diff
@@ -172,86 +172,95 @@ public boolean mutateGraph(Mutation mut) {
...
-        // adding a duplicate node doesn't make any change
-        if (!graphNodesMap.containsKey(startName)) {
-          // New lone node is a root
-          roots.add(startName);
-          // Create a new node with the given name and add it to the graph and the map
-          GraphNode newGraphNode =
-              GraphNode.create(startName, new ArrayList<>(), Struct.newBuilder().build());
-          graph.addNode(newGraphNode);
-          graphNodesMap.put(startName, newGraphNode);
+        {
```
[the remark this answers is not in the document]
haoyxc: ah i think this is a Google Java Format thing haha. I changed it and it has been changed back

src/main/java/com/google/sps/DataGraph.java
IshaniSanturkar: Oops @EatingW I misread the location of the loop, my response was in reference to another switch case. @haoyxc I think Yiting meant in the for-each loop like `for (GraphNode succ : graph.successors(startNode))`?

PR #27, opened 2020-07-14
Implement search on the backend

haoyxc:
- Server side: create a function that takes in a string name and radius, outputs a new graph with only nodes (successors and parents of that node) within the specified radius of the given node. Add corresponding tests
- Client side: add a search box for node name, pass that to the server (also improve the style of the boxes a little!)

Screenshot 2020-07-14 at 10 02 03 AM
Screenshot 2020-07-14 at 10 01 25 AM

Future optimizations:
- Right now the function of getting reachable nodes runs every time the search box changes. Possibly caching these graphs would prevent redundant work.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,7 +67,8 @@
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <version>4.13</version>
-      <!-- <scope>test</scope> -->
+      <!-- The following line is commented out because the tests weren't working for me with it -->
+      <!-- <scope>test</scope> --> 
     </dependency>
 
     <dependency>
--- a/src/main/java/com/google/sps/DataGraph.java
+++ b/src/main/java/com/google/sps/DataGraph.java
@@ -21,6 +21,7 @@ import com.proto.GraphProtos.Node;
 import com.proto.MutationProtos.Mutation;
 import com.proto.MutationProtos.TokenMutation;
 
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Map;
@@ -37,28 +38,22 @@ abstract class DataGraph {
    * @return the empty data graph with these attributes
    */
   public static DataGraph create() {
-    return new AutoValue_DataGraph(
-        GraphBuilder.directed().build(),
-        new HashMap<String, GraphNode>(),
-        new HashSet<String>(),
-        0);
+    return new AutoValue_DataGraph(GraphBuilder.directed().build(), new HashMap<String, GraphNode>(),
+        new HashSet<String>(), 0);
   }
 
   /**
    * Create a new data graph with the given attributes
    *
-   * @param graph the guava graph
+   * @param graph         the guava graph
    * @param graphNodesMap the map from node name to node
-   * @param roots a set of roots (nodes with no in-edges) of the graph
-   * @param numMutations the maximum number of mutations in the list of mutations this graph is an
-   *     intermediate result of applyiing
+   * @param roots         a set of roots (nodes with no in-edges) of the graph
+   * @param numMutations  the maximum number of mutations in the list of mutations
+   *                      this graph is an intermediate result of applyiing
    * @return the data graph with these attributes
    */
-  static DataGraph create(
-      MutableGraph<GraphNode> graph,
-      HashMap<String, GraphNode> graphNodesMap,
-      HashSet<String> roots,
-      int numMutations) {
+  static DataGraph create(MutableGraph<GraphNode> graph, HashMap<String, GraphNode> graphNodesMap,
+      HashSet<String> roots, int numMutations) {
     return new AutoValue_DataGraph(graph, graphNodesMap, roots, numMutations);
   }
 
@@ -111,10 +106,11 @@ abstract class DataGraph {
   }
 
   /**
-   * Takes in a map from node name to proto-parsed node object. Populates data graph with
-   * information from the parsed graph
+   * Takes in a map from node name to proto-parsed node object. Populates data
+   * graph with information from the parsed graph
    *
-   * @param protoNodesMap map from node name to proto Node object parsed from input
+   * @param protoNodesMap map from node name to proto Node object parsed from
+   *                      input
    * @return false if an error occurred, true otherwise
    */
   boolean graphFromProtoNodes(Map<String, Node> protoNodesMap) {
@@ -183,8 +179,7 @@ abstract class DataGraph {
           // New lone node is a root
           roots.add(startName);
           // Create a new node with the given name and add it to the graph and the map
-          GraphNode newGraphNode =
-              GraphNode.create(startName, new ArrayList<>(), Struct.newBuilder().build());
+          GraphNode newGraphNode = GraphNode.create(startName, new ArrayList<>(), Struct.newBuilder().build());
           graph.addNode(newGraphNode);
           graphNodesMap.put(startName, newGraphNode);
         }
@@ -254,10 +249,12 @@ abstract class DataGraph {
     }
     return true;
   }
+
   /**
-   * Modifies the list of tokens of this node to either add or remove tokens contained in tokenMut
+   * Modifies the list of tokens of this node to either add or remove tokens
+   * contained in tokenMut
    *
-   * @param node the node whose token list should be modified
+   * @param node     the node whose token list should be modified
    * @param tokenMut the mutation that should be applied to the token list
    * @return the new GraphNode object, or null if it's an unrecognized mutation
    */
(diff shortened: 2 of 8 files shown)


src/main/java/com/google/sps/DataGraph.java
```diff
@@ -327,4 +329,69 @@ private void dfsVisit(GraphNode gn, Map<GraphNode, Boolean> visited, int depthRe
       }
     }
   }
+
+  /**
+   * Returns a MutableGraph with nodes that are at most a certain radius from a given node If the
+   * radius is less than 0 or the node specified isn't present, return an emptu graph Since maxDepth
```
IshaniSanturkar, line 335: Small typo, change emptu to empty.

src/main/java/com/google/sps/DataGraph.java
```diff
@@ -327,4 +329,69 @@ private void dfsVisit(GraphNode gn, Map<GraphNode, Boolean> visited, int depthRe
...
+        if (!visited.containsKey(curr)) {
+          visited.put(curr, true);
+
+          // Adds the children
+          for (GraphNode child : graph.successors(curr)) {
+            if (!visited.containsKey(child)) {
+              nextLayer.add(child);
+            }
+          }
+          // Adds the parents
```
IshaniSanturkar, line 383: I thought we decided in poker planning that we would only return all reachable children and ancestors up to the root. Here since the parents and children are processed in one loop, it seems like we would also add other children of ancestors.

src/main/java/com/google/sps/DataGraph.java
```diff
@@ -327,4 +329,69 @@ private void dfsVisit(GraphNode gn, Map<GraphNode, Boolean> visited, int depthRe
...
   }
+
+  /**
+   * Returns a MutableGraph with nodes that are at most a certain radius from a given node If the
+   * radius is less than 0 or the node specified isn't present, return an emptu graph Since maxDepth
+   * was implemented as DFS, we use BFS for *diversity*
+   *
+   * @param name the name of the node to search for
+   * @param radius the distance from the node to search for parents and children
+   * @return a graph comprised of only nodes and edges within a certain distance from the specified
```
IshaniSanturkar, line 340: Indicate the return value in special cases like node not found etc.

src/main/webapp/script.js
```diff
@@ -145,7 +147,7 @@ function getUrl() {
       selectedDepth = 20;
     }
   }
-  const url = `/data?depth=${selectedDepth}&mutationNum=${currGraphNum}`;
+  const url = `/data?depth=${selectedDepth}&mutationNum=${currGraphNum}&nodeName=${nodeName}`;
```
IshaniSanturkar, line 150: Add some front-end tests for this?

src/main/java/com/google/sps/DataGraph.java
haoyxc: HAHA shoot

src/main/java/com/google/sps/DataGraph.java
haoyxc: ohhh ok my bad! so if you have B<--A-->C and looked for b, then it would just be A (and not C)?

haoyxc: ok changed!

haoyxc:
Updated based on feedback: 
image

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -103,7 +104,7 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
       // InputStreamReader(getServletContext().getResourceAsStream("/WEB-INF/mutations.textproto"));
       // MutationList.Builder mutBuilder = MutationList.newBuilder();
       // TextFormat.merge(mutReader, mutBuilder);
-      // List<Mutation> mutList = mutBuilder.build().getMutationList();
```
IshaniSanturkar, line 106: Merge with master, I think this code was uncommented.

src/main/java/com/google/sps/ReachableNodesTest.java
```diff
@@ -0,0 +1,258 @@
...
+import com.proto.GraphProtos.Node.Builder;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class ReachableNodesTest {
```
EatingW, line 18: I think these are all great tests! You could also add a test similar to the one you have in MaxDepthTest.java where a node is a reachable parent/child through one path but not another

src/main/java/com/google/sps/DataServlet.java
haoyxc: fixed!

pom.xml
```diff
@@ -67,7 +67,7 @@
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <version>4.13</version>
-      <scope>test</scope>
+      <!-- <scope>test</scope> -->
```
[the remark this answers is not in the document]
haoyxc: for some reason my tests we're working when I had that line. It told me the junit package wasn't found, so I googled for a bit and someone said to try it commented out and it worked (I didn't delete it because I'm not sure if this was a personal issue and wanted to leave it there in case tests didnt work for someone else)

src/main/java/com/google/sps/DataGraph.java
```diff
@@ -334,4 +335,107 @@ private void dfsVisit(GraphNode gn, Map<GraphNode, Boolean> visited, int depthRe
...
+        if (isChild) {
+          // Adds the children
+          for (GraphNode child : this.graph().successors(curr)) {
+            if (!visited.contains(child)) {
+              nextLayer.add(child);
+            }
+          }
+        } else {
+          // Adds the parents
+          for (GraphNode parent : graph().predecessors(curr)) {
```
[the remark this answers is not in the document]
haoyxc: got it! just wondering though, is this a style thing or because it helps the compiler (or something else?)

src/main/webapp/script.js
```diff
@@ -128,8 +128,10 @@ async function generateGraph() {
  */
 function getUrl() {
   const depthElem = document.getElementById('num-layers');
+  const nodeName = document.getElementById('node-name') ? document.getElementById('node-name').value || "" : "";
```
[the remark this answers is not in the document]
haoyxc: honestly it really shouldn't because it's in the html. I think I just did that because .value on something that's undefined doesn't sit well with JS, but do you think it's unnecessary?

src/main/java/com/google/sps/DataGraph.java
```diff
@@ -334,4 +335,107 @@ private void dfsVisit(GraphNode gn, Map<GraphNode, Boolean> visited, int depthRe
...
+      // Break out early if queue is empty
+      if (queueChildren.isEmpty() && queueParents.isEmpty()) {
+        break;
+      }
+
+      // Helper function used to avoid duplicate code
+      nextLayerChildren = getNextLayer(visitedChildren, queueChildren, true);
+      nextLayerParents = getNextLayer(visitedParents, queueParents, false);
+
+      queueChildren.addAll(nextLayerChildren);
```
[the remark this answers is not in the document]
haoyxc: oohh that's so smart!! thank you!! I don't know why but when I think BFS the first thing I thought of was a queue. Ok that was really cool actually & i changed it :)

src/main/java/com/google/sps/ReachableNodesTest.java
haoyxc: good idea!

src/main/java/com/google/sps/DataGraph.java
```diff
@@ -334,4 +335,107 @@ private void dfsVisit(GraphNode gn, Map<GraphNode, Boolean> visited, int depthRe
...
+  private HashSet<GraphNode> getNextLayer(
+      HashSet<GraphNode> visited, ArrayDeque<GraphNode> queue, boolean isChild) {
+    HashSet<GraphNode> nextLayer = new HashSet<>();
+    while (!queue.isEmpty()) {
+      GraphNode curr = queue.poll();
+
+      if (!visited.contains(curr)) {
+        visited.add(curr);
+
+        if (isChild) {
```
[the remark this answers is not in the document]
haoyxc: yes thank you!

pom.xml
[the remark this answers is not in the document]
haoyxc: I added a comment just in case to clarify things!

src/main/java/com/google/sps/DataGraph.java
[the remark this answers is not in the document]
EatingW: I think it's mostly a style thing.

src/main/webapp/script.js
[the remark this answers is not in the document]
EatingW: Directly calling `.value` would give a better indication of where the error is than if someone accidentally changed the id 'node-name' to something else like 'node-id', right? They would have a hard time tracking down the bug because the failure is silently converted to the default empty string.

EatingW:
Huh if I directly reply to an unresolved comment above it won't show up with my review. I should remember to use the review functionality to reply to future comments :)

PR #31, opened 2020-07-15
Search and highlight graph by token

connorbrett:
image

Enter token (file name) in search bar and press the button. The window will center/zoom on a node that contains the token and make other nodes transparent to focus on the selected node(s).


src/main/webapp/script.js
```diff
@@ -217,54 +217,105 @@ function getGraphDisplay(graphNodes, graphEdges) {
...
+    edge.style('line-color', 'black');
+    edge.style('target-arrow-color', 'black');
+  });
+}
+
+/**
+ * Resets elements to default state
+ */
+function resetElements(cy) {
+  cy.nodes().forEach(node => {
```
IshaniSanturkar, line 308: We might not want to do this for all nodes, for example if combined with the mutation highlighting feature. Could we just reset the elements we modified?

src/main/webapp/script.js
```diff
@@ -217,54 +217,105 @@ function getGraphDisplay(graphNodes, graphEdges) {
...
+    node.style('border-width', '0px');
+    node.style('opacity', '1')
+  });
+  cy.fit(cy.nodes(), 50);
+  document.getElementById('num-selected').innerText = "Number of nodes selected: 0";
```
IshaniSanturkar, line 313: Would we also want to reset the input field(s) for node name/token name?

src/main/webapp/script.js
```diff
@@ -217,54 +217,105 @@ function getGraphDisplay(graphNodes, graphEdges) {
...
+  // highlight desired nodes
+  target.forEach(node => {
+    node.style('border-width', '4px');
+    node.style('opacity', '1');
+  });
+  cy.fit(target[0], 50);
+  document.getElementById('num-selected').innerText = "Number of nodes selected: " + target.length;
+
+  // highlight adjacent edges
+  target.connectedEdges().forEach(edge => {
```
IshaniSanturkar, line 298: This is nice, but it makes it hard to distinguish the edges for a single node. Would it be possible to somehow cycle through the nodes in target, highlighting only that node and its edges at one time? Also, I don't know if this is a problem but what if the edge is covered by other edges. Is there a way to bring it to the forefront somehow?

src/main/webapp/script.js
```diff
@@ -217,54 +217,105 @@ function getGraphDisplay(graphNodes, graphEdges) {
...
+}
+/**
+ * Constructs list of nodes that contain specified token
+ * and zooms in
+ */
+function searchToken(cy, query) {
+  let target = cy.collection();
+  cy.nodes().forEach(node => {
+    if (node.data().tokens.includes(query)) {
+      target = target.add(cy.$id(node.data().id));
```
IshaniSanturkar, line 270: Isn't `cy.$id(node.data().id)` just `node`?

src/main/webapp/script.js
```diff
@@ -217,54 +217,105 @@ function getGraphDisplay(graphNodes, graphEdges) {
...
+    }
+  });
+  if (target.length > 0) {
+    let showNode = target[0][0];
+    showNode.tip.show();
+    highlightElements(cy, target);
+    return target;
+  }
+  //return false;
```
IshaniSanturkar, line 279: Remove the comment, and are you missing a return statement here?

src/main/webapp/script.js
```diff
@@ -217,54 +217,105 @@ function getGraphDisplay(graphNodes, graphEdges) {
...
-    target.style('background-color', 'olive');
-    target.style('opacity', '1');
-    cy.fit(target, 50);
-    return true;
+function search(cy, type, searchFunction) {
+  resetElements(cy);
+  let errorText;
+  let query = document.getElementById(type + '-search').value;
+  let result = searchFunction(cy, query);
+  if (query == "" || result) {
```
IshaniSanturkar, line 239: Since searching for the empty string is not allowed, would it be better to put the `if (query == "")` test before we call the search function to save some work?

src/main/webapp/index.html
```diff
@@ -22,6 +22,18 @@ <h1>
   <button onclick="graph.generateGraph()">Get Graph</button>
   <br />
 
+  <label>Find Node in Graph: </label>
+  <input type="text" id="node-search" placeholder="Node Name"></input>
+  <button id="search-button">Find Node</button>
+  <label id="node-error"></label>
+  <label>Find Token in Graph: </label>
+  <input type="text" id="token-search" placeholder="Token Name"></input>
+  <button id="search-token-button">Find Token</button>
```
IshaniSanturkar, line 31: Maybe insert some linebreaks between the label-input pairs? It looks a bit crowded when all the fields are populated.

spec/jstests-spec.js
```diff
@@ -305,41 +304,126 @@ describe("Check correct url params", function() {
...
-  it("should be an unsuccessful search", function() {
-    let result = searchNode(cy, "C");
+  it("should be an unsuccessful token search", function() {
+    query.value = "fake_file.js";
+    const result = search(cy, "token", searchToken);
 
-    // search should not find node
-    expect(result).toBe(false);
+    // error message should be displayed
+    expect(tokenError.innerText).not.toBe("");
```
IshaniSanturkar, line 416: Assert tokenError text.

spec/jstests-spec.js
```diff
@@ -305,41 +304,126 @@ describe("Check correct url params", function() {
...
+
+  it("should be a successful token search", function() {
+    query.value = "a.js";
+    const result = search(cy, "token", searchToken);
 
-    // search should find node
-    expect(result).toBe(true);
+    // error message should not be displayed
+    expect(tokenError.innerText).toBe("");
+    expect(result.length).toBe(1);
```
IshaniSanturkar, line 408: Make assertions about the content as well, that it is what we expected it to be. Also make sure that all the edges and nodes are highlighted the way we want them to be.

spec/jstests-spec.js
```diff
@@ -305,41 +304,126 @@ describe("Check correct url params", function() {
...
+
+    tokenError = document.createElement("label");
+    tokenError.id = "token-error";
+    document.body.appendChild(tokenError);
+
+    query = document.createElement("input");
+    query.id = "token-search";
+    document.body.appendChild(query);
+  });
+
```
IshaniSanturkar, line 400: Remove extra space.

spec/jstests-spec.js
```diff
@@ -305,41 +304,126 @@ describe("Check correct url params", function() {
...
+    nodeError = document.createElement("label");
+    nodeError.id = "node-error";
+    document.body.appendChild(nodeError);
+
+    query = document.createElement("input");
+    query.id = "node-search";
+    document.body.appendChild(query);
+  });
+
```
IshaniSanturkar, line 339: Extra empty line.

spec/jstests-spec.js
```diff
@@ -305,41 +304,126 @@ describe("Check correct url params", function() {
...
+    tokenError.id = "token-error";
+    document.body.appendChild(tokenError);
+
+    query = document.createElement("input");
+    query.id = "token-search";
+    document.body.appendChild(query);
+  });
+
+
+  it("should be a successful token search", function() {
```
IshaniSanturkar, line 402: Also do a case where there are multiple possible nodes.

spec/jstests-spec.js
```diff
@@ -305,41 +304,126 @@ describe("Check correct url params", function() {
...
+    document.body.appendChild(query);
+  });
+
+
+  it("should be a successful node search", function() {
+    query.value = "A";
+    const result = search(cy, "node", searchNode);
+
+    // should not display error message
+    expect(nodeError.innerText).toBe("");
```
IshaniSanturkar, line 346: Assert that the result is actually A.

spec/jstests-spec.js
```diff
@@ -305,41 +304,126 @@ describe("Check correct url params", function() {
...
+      ],
+      container: document.getElementById("cy"),
+    });
+    const nodeWithToken = {};
+    nodeWithToken["data"] = {};
+    nodeWithToken["data"]["id"] = "A";
+    nodeWithToken["data"]["tokens"] = ["a.js", "b.js", "c.js"];
+    cy.add(nodeWithToken);
+    const myNode = cy.nodes()[0];
+    initializeTippy(myNode);
```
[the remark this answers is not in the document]
connorbrett: the tippy has to be initialized because the token search opens the tip of a node that contains the token

spec/jstests-spec.js
```diff
@@ -305,41 +304,126 @@ describe("Check correct url params", function() {
...
+    // should not display error message
+    expect(nodeError.innerText).toBe("");
+  });
+
+  it("should be an unsuccessful node search", function() {
+    query.value = "C";
+    const result = search(cy, "node", searchNode);
+
+    // should display error message
+    expect(nodeError.innerText).not.toBe("");
```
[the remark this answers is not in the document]
connorbrett: How would you assert what it should be? Its just supposed to have an error message of some kind.

src/main/webapp/script.js
connorbrett: sure but the mutations only change the color of the nodes right? I'm not changing the color here at all so the mutated nodes colors should not change

src/main/webapp/script.js
connorbrett: For the first suggestion, yes I think so but that would be another feature entirely IMO and would likely require another PR. For the second suggestion, yes I think there is a way to change the z-index which would do what you are suggesting.

src/main/webapp/index.html
```diff
@@ -22,6 +23,22 @@ <h1>
...
+  <input type="text" id="node-search" placeholder="Node Name"></input>
+  <button id="search-button">Find Node</button>
+  <label id="node-error"></label>
+  <br />
+
+  <label>Find Token in Graph: </label>
+  <input type="text" id="token-search" placeholder="Token Name"></input>
+  <button id="search-token-button">Find Token</button>
+  <label id="token-error"></label>
+  <br/>
```
haoyxc, line 36: have you considered putting the whole thing in a div rather than using `<br/>`?

src/main/webapp/script.js
```diff
@@ -220,51 +220,117 @@ function getGraphDisplay(graphNodes, graphEdges) {
...
+/**
+ * Resets highlighted elements to default state
+ */
+function resetElements(cy) {
+  // reset node borders and opacity
+  cy.nodes().forEach(node => {
+    node.style('border-width', '0px');
+    // only change opacity of nodes that were changed
+    // because of highlighting
+    if (node.style('opacity') == "0.24") {
```
haoyxc, line 316: consider putting all these style things "black, 0px, .24" in constant objects? Right now if you want to change it you have to change it everywhere which is not great. Like `const style = {someAttribute: "black";...}` and instead of using .24 using `style[someAttribute]`

src/main/webapp/script.js
```diff
@@ -220,51 +220,117 @@ function getGraphDisplay(graphNodes, graphEdges) {
...
-  }
-  return null;
+  });
+
+  // reset edge color
+  cy.edges().forEach(edge => {
+    // only change color of edges that were changed
+    // because of highlighting
+    if (edge.style('line-color') == 'rgb(0,0,0)') {
+      edge.style('line-color', '#ccc');
```
haoyxc, line 326: same thing for all of these style things

src/main/webapp/script.js
```diff
@@ -220,51 +220,117 @@ function getGraphDisplay(graphNodes, graphEdges) {
...
+  document.getElementById('search-token-button').onclick = function() { search(cy, "token", searchToken) };
+}
+
+/**
+ * Searches based on type (node or token)
+ */
+function search(cy, type, searchFunction) {
+  resetElements(cy);
+  let errorText = "";
+  let query = document.getElementById(type + '-search').value;
```
haoyxc, line 236: does this change? why let?

src/main/webapp/script.js
```diff
@@ -220,51 +220,117 @@ function getGraphDisplay(graphNodes, graphEdges) {
...
+  let result;
+  if (query != "") {
```
haoyxc, line 238: have you considered using `!==`

haoyxc: https://codeburst.io/javascript-double-equals-vs-triple-equals-61d4ce5a121a

src/main/webapp/script.js
```diff
@@ -220,51 +220,117 @@ function getGraphDisplay(graphNodes, graphEdges) {
...
+    result = searchFunction(cy, query);
+    if (result) {
+      errorText = "";
```
haoyxc, line 241: isn't this what errorText is right now? have you considered doing
`if (!result) { // what you have down there}`

src/main/webapp/script.js
```diff
@@ -220,51 +220,117 @@ function getGraphDisplay(graphNodes, graphEdges) {
...
-      document.getElementById('search-error').innerText = "";
+  document.getElementById('reset').onclick = function(){ resetElements(cy) };
+
+  document.getElementById('search-button').onclick = function() { search(cy, "node", searchNode) };
+
+  document.getElementById('search-token-button').onclick = function() { search(cy, "token", searchToken) };
+}
+
+/**
+ * Searches based on type (node or token)
```
haoyxc, line 231: add a more descriptive comment? like what's returned, what the params mean, what changes?

src/main/webapp/script.js
```diff
@@ -220,51 +220,117 @@ function getGraphDisplay(graphNodes, graphEdges) {
...
-  };
-  
+  }
+  document.getElementById(type + '-error').innerText = errorText;
+  return result;
 }
 
 /**
- * Zooms in on specific node
+ * Finds specific node and zooms in
```
haoyxc, line 251: similarly, add something a little more descriptive? like what parameters mean and what it returns

src/main/webapp/script.js
```diff
@@ -220,51 +220,117 @@ function getGraphDisplay(graphNodes, graphEdges) {
...
-    target.style('background-color', 'olive');
-    target.style('opacity', '1');
-    cy.fit(target, 50);
-    return true;
-  } else {
-    // fits all nodes on screen
-    cy.fit(cy.nodes(), 50);
-    return false;
+  if (target.length > 0) {
+    let showNode = target[0][0];
```
haoyxc, line 275: does this change?

spec/jstests-spec.js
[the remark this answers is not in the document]
EatingW: You could either assert that it is equal to "node does not exist" (which is what you set as the error message) or use the `toContain` method to assert a substring of the error message. 
This is going into the implementation details a bit but I think ultimately asserting that the error message is not only present but also correct is important.

src/main/webapp/script.js
```diff
@@ -220,51 +220,117 @@ function getGraphDisplay(graphNodes, graphEdges) {
...
     }
-  }
-  return null;
+  });
+
+  // reset edge color
+  cy.edges().forEach(edge => {
+    // only change color of edges that were changed
+    // because of highlighting
+    if (edge.style('line-color') == 'rgb(0,0,0)') {
```
haoyxc, line 325: is there a reason you define the color with rgb here and `#000` in other places?

haoyxc: also `===`

src/main/webapp/script.js
EatingW, line 231: Also document the return value and that this changes the highlighting of the found node.

src/main/webapp/index.html
```diff
@@ -22,6 +23,22 @@ <h1>
...
+  
+  <button id="reset">Reset</button>
```
haoyxc, line 38: did you consider using div with styling rather than a a break tag?

src/main/webapp/script.js
```diff
@@ -220,51 +220,117 @@ function getGraphDisplay(graphNodes, graphEdges) {
...
+
+  document.getElementById('search-button').onclick = function() { search(cy, "node", searchNode) };
+
+  document.getElementById('search-token-button').onclick = function() { search(cy, "token", searchToken) };
+}
+
+/**
+ * Searches based on type (node or token)
+ */
+function search(cy, type, searchFunction) {
```
EatingW, line 233: searchAndHighlight might be a better name?

src/main/webapp/index.html
```diff
@@ -22,6 +23,22 @@ <h1>
   <button onclick="graph.generateGraph()">Get Graph</button>
   <br />
 
+  <label>Find Node in Graph: </label>
```
haoyxc, line 26: same thing with the break tag

src/main/webapp/script.js
EatingW, line 251: Document the return value.

src/main/webapp/script.js
```diff
@@ -220,51 +220,117 @@ function getGraphDisplay(graphNodes, graphEdges) {
...
+      cy.fit(target, 50);
+      return target;
+    }
+  }
+}
+/**
+ * Constructs list of nodes that contain specified token
+ * and zooms in
+ */
+function searchToken(cy, query) {
```
EatingW, line 267: Same here, also document the return value.

spec/jstests-spec.js
```diff
@@ -305,41 +304,145 @@ describe("Check correct url params", function() {
...
+
+    // should not display error message
+    expect(nodeError.innerText).toBe("");
+    expect(result.id()).toBe("A");
+  });
+
+  it("should be an unsuccessful node search", function() {
+    query.value = "C";
+    const result = search(cy, "node", searchNode);
+
```
EatingW, line 352: Since you have the result (target) here, could you also assert that the node is correctly highlighted? e.g. `node.numericStyle('border-width')`

spec/jstests-spec.js
```diff
@@ -305,41 +304,145 @@ describe("Check correct url params", function() {
...
+  });
+
+  it("should be a successful token search", function() {
+    query.value = "a.js";
+    const result = search(cy, "token", searchToken);
+
+    // error message should not be displayed
+    expect(tokenError.innerText).toBe("");
+    expect(result.length).toBe(1);
+    expect(result[0].id()).toBe("A");
```
EatingW, line 416: Same here, also assert that the highlight is correct?

src/main/webapp/script.js
```diff
@@ -220,51 +220,117 @@ function getGraphDisplay(graphNodes, graphEdges) {
...
+    // only change color of edges that were changed
+    // because of highlighting
+    if (edge.style('line-color') == 'rgb(0,0,0)') {
+      edge.style('line-color', '#ccc');
+      edge.style('target-arrow-color', '#ccc');
+      edge.style('z-index', '1');
+    }
+  });
+
+  //cy.fit(cy.nodes(), 50);
```
EatingW, line 332: delete?

src/main/webapp/script.js
```diff
@@ -220,51 +220,117 @@ function getGraphDisplay(graphNodes, graphEdges) {
...
-    target.style('opacity', '1');
-    cy.fit(target, 50);
-    return true;
-  } else {
-    // fits all nodes on screen
-    cy.fit(cy.nodes(), 50);
-    return false;
+  if (target.length > 0) {
+    let showNode = target[0][0];
+    showNode.tip.show();
```
EatingW, line 276: design question: Should we show the token list when highlighting? This only shows that of the first node in the graph, which might block the view of other nodes and might not be the one the user actually want to see.
Maybe it is sufficient to only highlight the nodes (and not display token list) and if the user is interested in seeing the token list they can manually click on the node they are interested in?

spec/jstests-spec.js
```diff
@@ -337,29 +339,32 @@ describe("Node search", function() {
     document.body.appendChild(query);
   });
 
-  it("should be a successful node search", function() {
```
EatingW: I think keeping the "node search" is a more descriptive test description!

spec/jstests-spec.js
```diff
@@ -337,29 +339,32 @@ describe("Node search", function() {
...
     query.value = "A";
     const result = search(cy, "node", searchNode);
 
     // should not display error message
     expect(nodeError.innerText).toBe("");
     expect(result.id()).toBe("A");
+    expect(result.style("border-width")).toBe("4px");
   });
 
-  it("should be an unsuccessful node search", function() {
```
EatingW: same here, as well as the token search descriptions below.

spec/jstests-spec.js
connorbrett: when the tests run it automatically prepends the string from the describe() function, so when the tests are run it has the full name
image

EatingW: Oh interesting, I didn't know it prepends the describe() string as well. Good find!

src/main/webapp/script.js
```diff
@@ -533,39 +537,96 @@ function hideDiffs(cy, elems, deletedNodes, deletedEdges, addedNodes, addedEdges
...
+ * Returns node if it exists
  */
 function searchNode(cy, query) {
-  // reset nodes to default color
+  let target = cy.$id(query);
+  if (target.length != 0) {
+    cy.fit(target, 50);
+    return target;
+  }
+}
```
IshaniSanturkar, line 571: Nit: space between functions.

src/main/webapp/script.js
```diff
@@ -533,39 +537,96 @@ function hideDiffs(cy, elems, deletedNodes, deletedEdges, addedNodes, addedEdges
...
-    const target = cy.$('#' + id);
-    if (target.length != 0) {
-      return target;
+function highlightElements(cy, target) {
+  cy.nodes().forEach(node => node.toggleClass('non-highlighted', true));
+
+  // highlight desired nodes
+  target.forEach(node => {
+    node.style('border-width', '4px');
+    node.toggleClass('non-highlighted', false);
```
IshaniSanturkar, line 600: I'm kind of confused. Shouldn't the class name of highlighted nodes be 'highlighted' as opposed to 'non-highlighted'?

src/main/webapp/script.js
```diff
@@ -533,39 +537,96 @@ function hideDiffs(cy, elems, deletedNodes, deletedEdges, addedNodes, addedEdges
...
-  if (id.length != 0) {
-    const target = cy.$('#' + id);
-    if (target.length != 0) {
-      return target;
+function highlightElements(cy, target) {
+  cy.nodes().forEach(node => node.toggleClass('non-highlighted', true));
+
+  // highlight desired nodes
+  target.forEach(node => {
+    node.style('border-width', '4px');
```
IshaniSanturkar, line 599: Is there a reason why you didn't put this in the initial configuration above like when you defined the opacity of 'non-highlighted' elements to be 0.25?

src/main/webapp/script.js
```diff
@@ -533,39 +537,96 @@ function hideDiffs(cy, elems, deletedNodes, deletedEdges, addedNodes, addedEdges
...
+  });
+}
+
+/**
+ * Resets highlighted elements to default state
+ */
+function resetElements(cy) {
+  // reset node borders and opacity
+  cy.nodes().forEach(node => {
+    node.style('border-width', '0px');
```
IshaniSanturkar, line 618: Not sure why you need to explicitly set the styles anymore. You could just toggle the classes and let the predefined configuration supplied while creating cy take care of the styling.

src/main/webapp/script.js
```diff
@@ -533,39 +537,96 @@ function hideDiffs(cy, elems, deletedNodes, deletedEdges, addedNodes, addedEdges
...
+    // only change opacity of nodes that were changed
+    // because of highlighting (leave nodes changed due to
+    // mutation alone)
+    if (node.hasClass('non-highlighted')) {
```
IshaniSanturkar, line 622: Also, isn't the purpose of toggleClass(_, false) to remove the class if it is present?

src/main/webapp/script.js
```diff
@@ -533,39 +537,96 @@ function hideDiffs(cy, elems, deletedNodes, deletedEdges, addedNodes, addedEdges
...
+      node.toggleClass('non-highlighted', false);
     }
-  }
-  return null;
+  });
+
+  // reset edge style
+  cy.edges().forEach(edge => edge.style('line-style', 'solid'));
```
IshaniSanturkar, line 628: Same for edges, you could just define a 'regular' edge style and a 'highlighted' edge style and use that here?

src/main/webapp/index.html
```diff
@@ -20,6 +20,23 @@ <h1>
     <input type="text" id="node-name" name="search-name" placeholder="Name"></input>
   </div>
   <button onclick="graph.generateGraph()">Get Graph</button>
+
+  <div>
+    <label>Find Node in Graph: </label>
```
haoyxc, line 25: for clarity, maybe "Zoom in on node" and "Zoom in on token" to distinguish

src/main/webapp/index.html
```diff
@@ -20,6 +20,23 @@ <h1>
...
+  </div>
+
+  <div>
+    <label>Find Token in Graph: </label>
+    <input type="text" id="token-search" placeholder="Token Name"></input>
+    <button id="search-token-button">Find Token</button>
+    <label id="token-error"></label>
+  </div>
+  
+  <button id="reset">Reset</button>
```
haoyxc, line 38: clarify this resets the graph and not the input fields maybe

src/main/webapp/index.html
```diff
@@ -20,6 +20,23 @@ <h1>
     <input type="text" id="node-name" name="search-name" placeholder="Name"></input>
   </div>
   <button onclick="graph.generateGraph()">Get Graph</button>
+
+  <div>
+    <label>Find Node in Graph: </label>
+    <input type="text" id="node-search" placeholder="Node Name"></input>
+    <button id="search-button">Find Node</button>
+    <label id="node-error"></label>
```
haoyxc, line 28: change the color of the text? it doesn't seem like an error right now

src/main/webapp/index.html
```diff
@@ -20,6 +20,23 @@ <h1>
...
+  </div>
+
+  <div>
+    <label>Find Token in Graph: </label>
```
haoyxc, line 32: image
when I find F by node and then click "Find Token" with b.js, nothing is highlighted and I don't get an error with "Token doesn't exist" either. Do you know what's happening?

src/main/webapp/index.html
```diff
@@ -20,6 +20,23 @@ <h1>
...
+    <input type="text" id="token-search" placeholder="Token Name"></input>
```
haoyxc, line 33: also when you search by a token that appears in multiple, it only displays the tippy of the first one. do you think it might be better to either display all or none? or say if it's greater than a certain number than display none else all
image

haoyxc: or you could have arrows that navigate to the next node (out of 174) and display the tippy

src/main/webapp/script.js
```diff
@@ -221,6 +222,12 @@ function getGraphDisplay(graphNodes, graphEdges, mutList, reason) {
           'target-arrow-color': colorScheme["unmodifiedEdgeColor"],
           'target-arrow-shape': 'triangle',
           'curve-style': 'bezier'
+        },
+      },
+      {
+        selector: '.non-highlighted',
+        style: {
+          'opacity': '0.25'
```
haoyxc, line 230: also constants!!

src/main/webapp/script.js
```diff
@@ -533,39 +537,96 @@ function hideDiffs(cy, elems, deletedNodes, deletedEdges, addedNodes, addedEdges
 }
 
 /**
- * Zooms in on specific node
+ * Searches a cytoscape graph (cy) based on type (node or token)
+ * using a specified search function.
+ * Returns the result of the search.
```
haoyxc, line 542: for consistency with the rest of script, use @param for paramters maybe? :)

src/main/webapp/script.js
```diff
@@ -533,39 +537,96 @@ function hideDiffs(cy, elems, deletedNodes, deletedEdges, addedNodes, addedEdges
...
+      errorText = type + " does not exist.";
+    }
+  }
+  document.getElementById(type + '-error').innerText = errorText;
+  return result;
+}
+
+/**
+ * Finds specific node from query and zooms in on it.
+ * Returns node if it exists
```
haoyxc, line 563: same thing with consistency @param

src/main/webapp/index.html
connorbrett: I am not sure and could not recreate it myself. What did you do to get that to happen?

haoyxc: have you tried navigating forward / back a few times and then searching for different things on the same graph? if you can't it's fine it might be a rare error I guess

src/main/webapp/index.html
```diff
@@ -50,5 +67,4 @@ <h1>
     </div>
 </div>
```
IshaniSanturkar, line 68: Nit: indent this div

src/main/webapp/script.js
```diff
@@ -602,39 +606,110 @@ function hideDiffs(cy, elems, deletedNodes, deletedEdges, addedNodes, addedEdges
...
+
+/**
+ * Resets collection of nodes/edges to default state
+ *
+ * @param cy the graph that contains nodes/edges
+ */
+function resetElements(cy) {
+  // reset node borders and opacity
+  cy.nodes().forEach(node => {
+    node.style('border-width', '0px');
```
IshaniSanturkar, line 701: Just want to note that in another branch @haoyxc uses border-width as well to highlight "queried" nodes. We should remember to integrate this there.

src/main/webapp/script.js
```diff
@@ -602,39 +606,110 @@ function hideDiffs(cy, elems, deletedNodes, deletedEdges, addedNodes, addedEdges
...
+    // only change opacity of nodes that were changed
+    // because of highlighting (leave nodes changed due to
+    // mutation alone)
+    if (node.hasClass('non-highlighted')) {
+      node.toggleClass('non-highlighted', false);
+    }
+  });
+
+  // reset edge style
+  cy.edges().forEach(edge => edge.style('line-style', 'solid'));
```
IshaniSanturkar, line 711: Would you also want to reset the z-index?

PR #35, opened 2020-07-16
Allow filtering through next and previous buttons

haoyxc:
Allows user to see only mutations relevant to a searched node
- Return a list of the relevant mutation indices to the client rather than just the mutation number
- Allow the user to search for a node, and respect the radius of the search
- Display the number of applicable mutations on the page as well as its position in the overall mutation list
- Created a map from nodeName -> mutation indices for caching
- Add a pop of (bright) color to the searched node
- Added tests to the new functions
- If the user is at an unfiltered graph and filters the graph, then if 1) the node is in the graph, it'll just be a generic search. if 2) the node is not on the graph, the user will see that and have the option to continue clicking forward. It will be graphs that are after the current graph their on. gotta love time series

image

Video:
[DNE.zip](https://github.com/googleinterns/step49-2020/files/4934687/DNE.zip)
(the mutation file I have isn't updated, so sometimes the clicks don't generate a viewable mutation. it's because sometimes there's stuff like 
image in my mutation file)

This PR doesn’t address the entirety of the “Integrate filtering by node name across graphs with next/previous buttons” issue. What's left is integrating with mutation reasoning/highlighting

Things to consider
If user searches for a node and then clears, what graph should they be brought to? (the initial one, or the same one they’re currently on with all the nodes?) 
The current implementation is that if the filter changes, the graph will be the same. Just with different nodes.

note: please let me know any issues! I really wanted some feedback on implementation!

### DIFF
--- a/src/main/java/com/google/sps/DataServlet.java
+++ b/src/main/java/com/google/sps/DataServlet.java
@@ -130,7 +130,6 @@ public class DataServlet extends HttpServlet {
       return;
     }
 
-
     List<Mutation> truncatedMutList;
 
     MutableGraph<GraphNode> truncatedGraph;
@@ -142,16 +141,24 @@ public class DataServlet extends HttpServlet {
       truncatedGraph = currDataGraph.getGraphWithMaxDepth(depthNumber);
       truncatedMutList = mutList;
     } else {
-      // This is the single search
-      truncatedGraph = currDataGraph.getReachableNodes(nodeNameParam, depthNumber);
 
       // Indicies of relevant mutations
       relevantMutationIndices = Utility.getMutationIndicesOfNode(nodeNameParam, mutList);
+
+      // TODO: find the index that's the next greatest on this list with binary search
+      // That is, change the mutation num!!!
+      int newNum = Utility.getNextGreatestNum(relevantMutationIndices, oldNumMutations);
+
+      // Maybe make a copy instead of making this the currDataGraph
+      currDataGraph = Utility.getGraphAtMutationNumber(originalDataGraph, currDataGraph, newNum, mutList);
 
       // If the truncated graph is empty, it doesn't exist on the page. Check if there
       // are any
       // mutations that affect it
       truncatedMutList = Utility.getMutationsFromIndices(relevantMutationIndices, mutList); // only mutations relevant to the node
+
+      // This is the single search
+      truncatedGraph = currDataGraph.getReachableNodes(nodeNameParam, depthNumber);
 
       // oldNumMutations is the number of mutations that were applied.
       // you want to see where this falls in the new one
--- a/src/main/java/com/google/sps/DataServlet.java
+++ b/src/main/java/com/google/sps/DataServlet.java
@@ -37,6 +37,8 @@ public class DataServlet extends HttpServlet {
   private DataGraph currDataGraph = null;
   private DataGraph originalDataGraph = null;
 
+  private String latestSearchNode = null;
+
   /*
    * Called when a client submits a GET request to the /data URL
    */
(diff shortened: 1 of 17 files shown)


src/main/java/com/google/sps/NodeMutationFilterTest.java
```diff
@@ -0,0 +1,173 @@
...
+   * index is returned)
+   */
+  @Test
+  public void testInnerExistsEven() {
+    int ans = Utility.getNextGreatestNumIndex(lst1, 4);
+    Assert.assertEquals(2, ans);
+  }
+
+  @Test
+  public void testInnerExistsOdd() {
```
IshaniSanturkar, line 127: This test would make more sense if you "controlled" what you were searching for (like searching for 4 in both cases).

src/main/java/com/google/sps/NodeMutationFilterTest.java
```diff
@@ -0,0 +1,173 @@
...
+    Assert.assertEquals(3, ans);
+  }
+
+  /**
+   * When tgt value does not exist (DNE) in the list, value is just the immediate greater one. This
+   * tests a middle index is returned correctly.
+   */
+  @Test
+  public void testInnerDNEEven() {
+    int ans = Utility.getNextGreatestNumIndex(lst1, 1);
```
IshaniSanturkar, line 138: 1 does exist in lst1?

src/main/webapp/index.html
```diff
@@ -21,7 +21,7 @@ <h1>
   </div>
   <button onclick="graph.generateGraph()">Get Graph</button>
   <br />
-
+  <div id="num-mutation-display"></div>
```
IshaniSanturkar, line 24: I noticed in your demo that there were 2 graph divs. Could you delete one of them?

IshaniSanturkar:
Not done reviewing completely, I'll take some time to understand the code more!

src/main/webapp/script.js
```diff
@@ -351,14 +392,16 @@ function navigateGraph(amount) {
...
  */
 function updateButtons() {
-  if (currGraphNum === 0) {
+  if (currGraphIndex === 0) {
     document.getElementById("prevbutton").disabled = true;
   } else {
     document.getElementById("prevbutton").disabled = false;
   }
-  if (currGraphNum === numMutations) {
+  if (currGraphIndex === numMutations - 1|| numMutations === 0) {
```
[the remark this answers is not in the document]
haoyxc: we love gvc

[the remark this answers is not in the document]
haoyxc: [don't want an index that's out of bounds]

src/main/webapp/script.js
```diff
@@ -351,14 +392,16 @@ function navigateGraph(amount) {
...
     document.getElementById("nextbutton").disabled = true;
   } else {
     document.getElementById("nextbutton").disabled = false;
   }
+  const numElement = document.getElementById("num-mutation-display");
+  numElement.innerText = `Displaying ${currGraphIndex + 1} out of ${numMutations} (this is ${currGraphNum + 1} on the original!)`;
```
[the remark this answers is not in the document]
haoyxc: also think we resolved this one, fixed the !

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -99,37 +109,117 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-
       InputStreamReader mutReader =
           new InputStreamReader(
-              getServletContext().getResourceAsStream("/WEB-INF/mutation.textproto"));
+              getServletContext().getResourceAsStream("/WEB-INF/mutations.textproto"));
       MutationList.Builder mutBuilder = MutationList.newBuilder();
       TextFormat.merge(mutReader, mutBuilder);
       mutList = mutBuilder.build().getMutationList();
+
+      // initially, all mutation indices are relevant so we put that into the default
```
connorbrett, line 119: maybe document what relevant indices are somewhere because it can be confusing

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -99,37 +109,117 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
     }
 
+    // Truncated version of the graph and mutation list, for returning to the client
     MutableGraph<GraphNode> truncatedGraph;
 
-    // If a node is searched, get the graph with just the node. Otherwise, use the
-    // whole graph
+    // No node is searched, so use the whole graph
     if (nodeNameParam == null || nodeNameParam.length() == 0) {
+      // Just get the specified deptg, the mutation list, and relevant mutations as
```
connorbrett, line 147: small typo, depth*

PR #41, opened 2020-07-21
Integrate highlighting with filtering by node name 

haoyxc:
Changes in this PR include: 
- Start the numMuations in DataGraph as -1. numMuations refers to the last mutation index that has been applied to the graph. -1 means that there have been 0 applied. updated the frontend to match this.
- Add binary search for closest smallest & largest indices on the frontend
- Make the filtering compatible with the multi-mutation format, filter the multimutation by nodes so that only relevant mutations in the multimutation exist in the graph
- Create a separate file in JS for storing cytoscape constants
- And include things from the mutationreason & filterAcc branches: option to check/uncheck the highlighting, option to filter by a given node and see the mutations that are only relevant to the node
- Add a side panel in index to display messages (such as a node being in the graph but in future graphs, etc)
- simplify DataServlet logic based on cases discussed in poker planning (the backend always received the correct indices now, so there's simpler logic!)

I apologize in advanced this is bigger than I anticipated!! Since we're disregarding one of the previous PRs, this includes that stuff AND some more code!

Original:
Screenshot 2020-07-21 at 4 54 54 PM
Node does not exist:
Screenshot 2020-07-21 at 4 55 18 PM
Node exists, not in current: (message displayed on the side in red)
Screenshot 2020-07-21 at 4 55 34 PM
Node exists, in current graph: (changes highlighted)
Screenshot 2020-07-21 at 4 55 46 PM
Screenshot 2020-07-21 at 4 57 54 PM

While the core functionality is done, some of the code isn't great (and we're still working on cleaning it up!)


src/main/webapp/script.js
```diff
@@ -25,36 +25,34 @@ import tippy, { sticky } from 'tippy.js';
...
 
 cytoscape.use(popper); // register extension
 cytoscape.use(dagre); // register extension
 
-// Stores the index of the graph (in sequence of mutations) currently
-// displayed on the screen. Must be >= 0.
-let currGraphNum = 0;
-// Stores the number of mutations in the list this graph is applying
-// The user cannot click next to a graph beyond this point
+// Stores the index of the most recently-applied mutation in mutationIndexList
```
IshaniSanturkar, line 40: Note that this could be a decimal and explain what that means.

src/main/webapp/script.js
```diff
@@ -25,36 +25,34 @@ import tippy, { sticky } from 'tippy.js';
...
-  "modifiedNodeColor": "yellow",
-  "unmodifiedEdgeColor": "grey",
-  "labelColor": "white"
-};
-// Sets the opacity constants for different types of objects in the graph
-const opacityScheme = {
-  "deletedObjectOpacity": 0.25
-};
+
+// Stores the list of relevant mutation indices
```
IshaniSanturkar, line 52: More descriptive comment here.

src/main/webapp/script.js
```diff
@@ -64,10 +62,24 @@ function initializeNumMutations(num) {
...
 /** 
- * Sets the current graph number
+ * Sets the current mutation number
+ */
+function setCurrMutationNum(num) {
+  currMutationNum = num;
+}
+
+/**
+ * Sets the current graph index
```
IshaniSanturkar, line 72: Update comment.

src/main/webapp/script.js
```diff
@@ -64,10 +62,24 @@ function initializeNumMutations(num) {
...
+ */
+function setCurrMutationIndex(num) {
+  currMutationIndex = num;
+}
+
+/**
+ * Sets the relevant indices from the mutation list
```
IshaniSanturkar, line 79: Not sure what the comment means here.

src/main/webapp/script.js
```diff
@@ -104,19 +112,34 @@ async function generateGraph() {
...
   const edges = JSON.parse(jsonResponse.edges);
-  initializeNumMutations(JSON.parse(jsonResponse.numMutations));
+
   const mutList = jsonResponse["mutationDiff"].length === 0 ? null : JSON.parse(jsonResponse["mutationDiff"]);
   const reason = jsonResponse["reason"];
 
+  mutationIndexList = JSON.parse(jsonResponse.mutationIndices);
+  numMutations = mutationIndexList.length;
+
+  // currMutationIndex = jsonResponse.currIndex;
```
IshaniSanturkar, line 122: Remove commented out code.

src/main/webapp/script.js
```diff
@@ -104,19 +112,34 @@ async function generateGraph() {
...
-  if (nodes.length === 0) {
-    displayError("Nothing to display!");
+  // There aren't any nodes in this graph, and there aren't any mutations pertaining to the filtered node
+  if (nodes.length === 0 && numMutations === 0) {
+    displayError("This node does not exist in any stage of the graph!");
     return;
+  } else if (response.headers.get("serverMessage")) {
+    // This happens if the graph doesn't contain the searched node or 
+    // if the graph contains the searched node BUT it isn't mutated in this graph
+    // We have to adjust the indices in this case
```
IshaniSanturkar, line 137: Move this comment outside the else if case closer to the code below.

src/main/webapp/script.js
```diff
@@ -167,9 +190,19 @@ function getUrl() {
       selectedDepth = 20;
     }
   }
-  const url = `/data?depth=${selectedDepth}&mutationNum=${currGraphNum}&nodeName=${nodeName}`;
+  const url = `/data?depth=${selectedDepth}&mutationNum=${currMutationNum}&nodeName=${nodeName}`;
   return url;
 }
+
+/**
+ * Add to the TOP of the logs list
```
IshaniSanturkar, line 198: Add more documentation here.

src/main/webapp/script.js
```diff
@@ -636,15 +690,30 @@ function getTooltipContent(node) {
...
-  currGraphNum += amount;
-  if (currGraphNum <= 0) {
-    currGraphNum = 0;
+  // this function should not be called if there are no mutations
+  if (numMutations <= 0) {
+    return;
   }
-  if (currGraphNum >= numMutations) {
-    currGraphNum = numMutations;
+  if (Number.isInteger(currMutationIndex)) {
```
IshaniSanturkar, line 701: Explain what these cases mean.

src/main/webapp/script.js
```diff
@@ -654,14 +723,72 @@ function navigateGraph(amount) {
  * Assumes currGraphNum is between 0 and numMutations
  */
 function updateButtons() {
-  if (currGraphNum === 0) {
+  // The use of <= and >= as opposed to === is for safety! 
+  // while currGraphIndex should never be < 0 or > numMutations - 1, we just wanted to make sure
```
IshaniSanturkar, line 727: old comments that don't really make sense anymore.

src/main/webapp/script.js
```diff
@@ -654,14 +723,72 @@ function navigateGraph(amount) {
...
     document.getElementById("nextbutton").disabled = false;
   }
+  const numElement = document.getElementById("num-mutation-display");
+  numElement.innerText = `Graph ${currMutationNum + 1}`;
+}
+
+/**
+ * Get the index of the next largest element in an indiceList
+ * @param indicesList a list of indices, assume it's sorted
+ * @param element the element to find a higher value than
```
IshaniSanturkar, line 747: Document return value here and below.

src/main/java/com/google/sps/NodeMutationFilterTest.java
```diff
@@ -0,0 +1,181 @@
...
+    multiMutList.add(addABM);
+    multiMutList.add(removeABM);
+    multiMutList.add(removeCM);
+
+    List<Integer> truncatedList = Utility.getMutationIndicesOfNode("A", multiMutList);
+
+    Assert.assertEquals(2, truncatedList.size());
+    Assert.assertTrue(truncatedList.contains(0));
+    Assert.assertTrue(truncatedList.contains(1));
+    Assert.assertFalse(truncatedList.contains(2));
```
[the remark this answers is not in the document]
haoyxc: oh whoops I think i changed it on the previous PR

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -55,73 +119,111 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+
+    // Get the diff if we are going forward in the list of mutations
+    if (mutationNumber > currDataGraph.numMutations()) {
+      diff = Utility.getDiffBetween(mutList, mutationNumber);
     }
+    // Find the indices that mutate the searched node, computing and caching them
+    // if this has not been done already
+    if (!mutationIndicesMap.containsKey(nodeNameParam)) {
+      mutationIndicesMap.put(
+          nodeNameParam, Utility.getMutationIndicesOfNode(nodeNameParam, mutList));
```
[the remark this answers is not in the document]
haoyxc: the alternative would be to not store it. then if they search the same node that doesn't exist again, it would have to get the indices again which seems like not great either. 

What do you think?

src/main/webapp/script.js
```diff
@@ -25,36 +25,36 @@ import tippy, { sticky } from 'tippy.js';
 import 'tippy.js/dist/tippy.css';
 import 'tippy.js/dist/backdrop.css';
 import 'tippy.js/animations/shift-away.css';
+import { colorScheme, opacityScheme } from './constants.js';
 
 export {
-  searchNode, initializeNumMutations, setCurrGraphNum, initializeTippy,
-  generateGraph, getUrl, navigateGraph, currGraphNum, numMutations, updateButtons,
-  highlightDiff, initializeReasonTooltip, getGraphDisplay
+  initializeNumMutations, setMutationIndexList, setCurrMutationNum, initializeTippy,
```
connorbrett, line 31: is there any way to simplify this long list of functions or work around it like just exporting the whole file? at this point, it is getting difficult to read and use

haoyxc:
Screenshot 2020-07-22 at 4 42 04 PM

src/main/webapp/script.js
```diff
@@ -634,15 +690,36 @@ function getTooltipContent(node) {
  * When a next/previous button is clicked, modifies the mutation index of the
```
EatingW, line 696: Also add some high-level documentation of what happens when the currMutationIndex is decimal

src/main/webapp/script.js
```diff
@@ -634,15 +690,36 @@ function getTooltipContent(node) {
...
+  // this function should not be called if there are no mutations
+  if (numMutations <= 0) {
+    return;
+  }
+  if (Number.isInteger(currMutationIndex)) {
+    // In this case, currMutationNum is in mutationIndexList, so update the 
+    // index by the given amount (either +1 or -1)
+    currMutationIndex += amount;
+  } else {
+    // In this case, currMutationNum is the average of two adjacent indices
```
EatingW, line 712: Seems like there are no test cases for this scenario?

src/main/webapp/script.js
```diff
@@ -339,6 +388,8 @@ function highlightDiff(cy, mutList, reason = "") {
           // color this node green
           modifiedObj.style('background-color', colorScheme["addedObjectColor"]);
           addedNodes = addedNodes.union(modifiedObj);
+        } else {
+          addToLogs("No node called " + startNode + " in graph");
```
EatingW, line 398: When is this state possible? Shouldn't you already have filtered the mutations down to only the ones relevant to the nodes displayed? What does this convey to the user when they see it (i.e. is it actionable for the users)?

src/main/webapp/script.js
```diff
@@ -652,14 +729,74 @@ function navigateGraph(amount) {
...
     document.getElementById("prevbutton").disabled = false;
   }
-  if (currGraphNum === numMutations) {
+  if (currMutationIndex >= mutationIndexList.length - 1) {
     document.getElementById("nextbutton").disabled = true;
   } else {
     document.getElementById("nextbutton").disabled = false;
   }
+  const numElement = document.getElementById("num-mutation-display");
+  numElement.innerText = `Graph ${currMutationNum + 1}`;
```
EatingW, line 750: I think it might still be valuable to display "Graph {currMutation} of {totalMutation}"

src/main/webapp/script.js
```diff
@@ -652,14 +729,74 @@ function navigateGraph(amount) {
...
+
+/**
+ * Get the index of the next largest element in an indiceList
+ * @param indicesList a list of indices, assume it's sorted
+ * @param element the element to find a higher value than
+ * @return the index of the first element in indicesList larger than
+ * element or indicesList.length if element is larger than all elements
+ * in indicesList
+ */
+function getIndexOfNextLargerNumber(indicesList, element) {
```
EatingW, line 761: I wonder if you could merge `getIndexOfNextLargerNumber` and `getIndexOfPrevSmallerNumber` into the same function. Since you already have an index of the next larger, you could just do a traversal backwards to find the index of the first number smaller than element, right? It should be at most 2 traversals backwards, unless you have multiple elements in the list with the same value (e.g. [5,5,5,5,5] in your test case) but in practice this won't happen for this tool in the real world scenario :)

src/main/webapp/script.js
```diff
@@ -167,9 +191,26 @@ function getUrl() {
...
+ * @param msg the message to display in the new list element
+ */
+function addToLogs(msg) {
+  const logsList = document.getElementById("log-list");
+  const newMsg = document.createElement("li");
+  newMsg.classList.add("log-msg");
+  newMsg.innerText = msg;
+  logsList.appendChild(newMsg);
+  const allLogs = document.querySelectorAll('.log-msg');
+  for (let i = 0; i < allLogs.length; i++) {
```
EatingW, line 209: You only need to remove "error-text" for the very last element, right? Everytime you addToLogs only at most one log (the last one) should be red. 

Also I feel like whenever the user sends a new request to the server, we should set the most recent log message to black as well so that if the new request does not have an error, the logs shouldn't have anything red.

src/main/webapp/script.js
haoyxc: what do you suggest we could do?

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -57,86 +119,118 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    if (truncatedGraph.nodes().size() == 0 && filteredMutationIndices.size() == 0) {
+      response.setHeader(
+          "serverError", "The searched node does not exist anywhere in this graph or in mutations");
+      return;
+    }
+    // The searched node is not in the graph but is mutated at some past/future point
+    if (truncatedGraph.nodes().size() == 0 && filteredMutationIndices.size() != 0) {
+      response.setHeader(
+          "serverMessage",
+          "The searched node does not exist in this graph, but is mutated at some point");
```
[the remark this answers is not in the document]
haoyxc: we use it to set the value of success - do you think this is unnecessary?
image

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -57,86 +119,118 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+   *
+   * @param mutationInput InputStream to initialize variable over
+   * @return whether variables were initialized properly; true if successful and false otherwise
+   * @throws IOException if something goes wrong during the reading
+   */
+  private void initializeMutationVariables(InputStream mutationInput) throws IOException {
+    InputStreamReader mutReader = new InputStreamReader(mutationInput);
+    MutationList.Builder mutBuilder = MutationList.newBuilder();
+    TextFormat.merge(mutReader, mutBuilder);
+    mutList = mutBuilder.build().getMutationList();
```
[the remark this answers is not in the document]
haoyxc: we use it to set the value of success - do you think this is unnecessary?
image

src/main/java/com/google/sps/MutationDiffTest.java
```diff
@@ -317,7 +278,195 @@ public void forwardMutationDeleteNode() {
...
+        Mutation.newBuilder()
+            .setType(Mutation.Type.DELETE_EDGE)
+            .setStartNode("A")
+            .setEndNode("B")
+            .build();
+    Mutation removeB =
+        Mutation.newBuilder().setType(Mutation.Type.DELETE_NODE).setStartNode("B").build();
+    MultiMutation originalMultiMut =
+        MultiMutation.newBuilder()
+            .addMutation(addTokenToA)
```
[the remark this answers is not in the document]
haoyxc: it would also contain addTokenToB I believe! Will add a test

src/main/java/com/google/sps/NodeMutationFilterTest.java
```diff
@@ -0,0 +1,91 @@
...
+    MultiMutation removeCM = MultiMutation.newBuilder().addMutation(removeC).build();
+    List<MultiMutation> multiMutList = new ArrayList<>();
+    multiMutList.add(addABM);
+    multiMutList.add(removeABM);
+    multiMutList.add(removeCM);
+
+    List<Integer> truncatedList = Utility.getMutationIndicesOfNode("A", multiMutList);
+
+    Assert.assertEquals(2, truncatedList.size());
+    Assert.assertTrue(truncatedList.contains(0));
```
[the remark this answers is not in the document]
haoyxc: I was getting `reference to assertEquals is ambiguous` with assertEquals, so I just changed it to a assertTrue with an equals! Hope it's okay. I think I read that you could also convert it to a long value? I'm not sure if that would be better, do you have a preference?

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -57,86 +119,118 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+      response.setHeader(
+          "serverMessage",
+          "The searched node does not exist in this graph, but is mutated at some point");
+    }
+    // The searched node exists but is not mutated in the current graph
+    if (truncatedGraph.nodes().size() != 0
+        && mutationNumber != -1
+        && filteredMutationIndices.indexOf(mutationNumber) == -1) {
+      response.setHeader(
+          "serverMessage", "The searched node exists, but is not mutated in this graph");
```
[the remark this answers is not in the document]
haoyxc: do you think it's useful to display, or should we just take it out?

src/main/webapp/script.js
haoyxc: I think it's in line 41! do you think we should move it somewhere else?

src/main/webapp/script.js
haoyxc: for the total, should that be the total number of relevant mutations there are? because if so, would it be weird if it said like `Graph 29 out of 10` (this would happen if there were 10 mutations relevant to a node searched, but it was graph 29 overall since we're displaying the mutationNum rather than index).

src/main/webapp/script.js
haoyxc: think i fixed it! let me know if there are problems with it!

src/main/webapp/script.js
haoyxc: I just added something with the total number of graphs in parenthesis! please follow up if something else would be better!

src/main/java/com/google/sps/DataServlet.java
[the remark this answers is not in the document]
EatingW: The one I'm referring to is `initializeMutationVariables` though, which has a function signature of `void` but the JavaDoc comment says @return a boolean :p

spec/jstests-spec.js
```diff
@@ -209,12 +209,15 @@ describe("Pressing next and previous buttons associated with a graph", function
 
   afterEach(function () {
     document.body.innerHTML = '';
+    // setCurrMutationIndex(-1);
```
EatingW, line 212: delete?

spec/jstests-spec.js
```diff
@@ -290,12 +294,36 @@ describe("Pressing next and previous buttons associated with a graph", function
     expect(currMutationIndex).toBe(-1);
     expect(nextButton.disabled).toBe(false);
     expect(prevButton.disabled).toBe(true);
+  });
+  it("Navigate graph returns with no mutations doesn't change anything", function() {
```
EatingW, line 298: nit: newline between tests

spec/jstests-spec.js
```diff
@@ -290,12 +294,36 @@ describe("Pressing next and previous buttons associated with a graph", function
...
+   // Nothing changes with a negative numMutations
+    navigateGraph(-1);
+    expect(currMutationNum).toBe(10);
 
-    nextButton.click();
-    expect(currMutationNum).toBe(0);
+    navigateGraph(1);
+    expect(currMutationNum).toBe(10);
+  });
+  it("Index is a decimal navigate forward", function() {
```
EatingW, line 310: nit: the description for the `it` function is appended to the description of `describe` "Pressing next and previous buttons associated with a graph" when tests are run, so its generally a convention and helps readability to write the description for each test such that it connects to a full sentence.
e.g. `it("correctly navigates forward when index is a decimal")` => "Pressing next and previous buttons associated with a graph correctly navigates forward when index is a decimal"

src/main/webapp/script.js
EatingW: line 41 talks about the fact that currMutationIndex could be decimal. What could be added here is what happens when currMutationIndex is decimal. e.g. "if currMutationIndex is decimal, next/prev would set it to the next/previous integer index"

src/main/webapp/script.js
EatingW: I was thinking of displaying for example `Graph 29 out of 3000` where the total number of mutations is displayed. To do this you'll have to send an additional piece of information over from the server.

src/main/webapp/script.js
```diff
@@ -756,62 +756,42 @@ function updateButtons() {
...
-  numElement.innerText = `Graph ${currMutationNum + 1} out of ${totalNum} total!)`;
+  numElement.innerText = `Graph ${currMutationNum + 1} out of ${totalNum} total`;
 }
-
+ 
 /**
- * Get the index of the next largest element in an indiceList
+ * Get an object with the index of the element that's immediately smaller 
+ * and immediately greater than the element
  * @param indicesList a list of indices, assume it's sorted
```
EatingW, line 765: nit: add the requirement that there should be no duplicates.

PR #42, opened 2020-07-22
Allow filtering by token in a single graph

haoyxc:
Changes in this PR:
- Generalize reachable nodes function to take a Collection of names, names will have “” which is the empty string all of the time no node is searched in the node name field. If “” is the only thing in names, we return an empty string
- Add test cases for Collection functionality
- Add input field for token on client
- Add tokenMap field to the DataGraph, update tokenMap to reflect mutations, modified test cases to check for these changes

image
Sanity check: (there's 10 nodes with this token in the initial)
image


src/main/java/com/google/sps/DataGraph.java
```diff
@@ -295,8 +322,33 @@ private GraphNode changeNodeToken(GraphNode node, TokenMutation tokenMut) {
...
       tokenList.removeAll(tokenNames);
+      // Update the map
+      for (String tokenName : tokenNames) {
+        if (this.tokenMap().containsKey(tokenName)) {
+          Set<String> nodesWithToken = this.tokenMap().get(tokenName);
+          nodesWithToken.remove(node.name());
+          if (nodesWithToken.size() == 0) { // No more nodes with token
+            this.tokenMap().remove(tokenName);
+          } else {
+            this.tokenMap().put(tokenName, nodesWithToken);
```
connorbrett, line 348: not sure where to put this but when you filter by token, its not immediately clear which nodes have said token, in my searchToken branch i have some functions that handle highlighting by token and resetting them if you want to look at that

src/main/java/com/google/sps/ReachableNodesTest.java
```diff
@@ -297,4 +308,167 @@ public void findsShorterPath() {
...
+    HashMap<String, Node> protoNodesMap = new HashMap<>();
+    protoNodesMap.put("A", nodeA.build());
+    protoNodesMap.put("B", nodeB.build());
+    protoNodesMap.put("C", nodeC.build());
+
+    DataGraph dataGraph = DataGraph.create();
+    dataGraph.graphFromProtoNodes(protoNodesMap);
+    List<String> lst = new ArrayList<>(Arrays.asList(""));
+
+    MutableGraph<GraphNode> truncatedGraph = dataGraph.getReachableNodes(lst, 2);
```
IshaniSanturkar, line 409: Could we try this with a parameter of 0 maybe? It could be the case that for empty string, the function just returns the whole graph.

src/main/java/com/google/sps/ReachableNodesTest.java
```diff
@@ -297,4 +308,167 @@ public void findsShorterPath() {
...
+    DataGraph dataGraph = DataGraph.create();
+    dataGraph.graphFromProtoNodes(protoNodesMap);
+    List<String> lst = new ArrayList<>(Arrays.asList("", "D"));
+
+    MutableGraph<GraphNode> truncatedGraph = dataGraph.getReachableNodes(lst, 2);
+    Set<GraphNode> graphNodes = truncatedGraph.nodes();
+    Set<EndpointPair<GraphNode>> graphEdges = truncatedGraph.edges();
+
+    Assert.assertEquals(0, graphNodes.size());
+    Assert.assertFalse(graphNodes.contains(gNodeA));
```
IshaniSanturkar, line 441: Redundant checks?

src/main/java/com/google/sps/DataGraph.java
```diff
@@ -107,9 +117,14 @@ public DataGraph getCopy() {
     for (String key : graphNodesMap.keySet()) {
       copyMap.put(key, graphNodesMap.get(key));
     }
+
+    HashMap<String, Set<String>> tokenCopyMap = new HashMap<>();
+    for (String key : this.tokenMap().keySet()) {
+      tokenCopyMap.put(key, this.tokenMap().get(key));
```
IshaniSanturkar, line 123: Since we want our internal data structures to be "immutable" for a data graph, when we create a copy of this map should we also create a copy of the set for each token. I know that in add/remove token we modify this set, which would result in earlier versions possibly being modified as well.

src/main/webapp/script.js
```diff
@@ -202,7 +203,7 @@ function getUrl() {
       selectedDepth = 20;
     }
   }
-  const url = `/data?depth=${selectedDepth}&mutationNum=${currMutationNum}&nodeName=${nodeName}`;
+  const url = `/data?depth=${selectedDepth}&mutationNum=${currMutationNum}&nodeName=${nodeName}&tokenName=${tokenName}`;
```
IshaniSanturkar, line 206: Nit: add a check in the front-end tests for new parameter.

src/main/webapp/script.js
```diff
@@ -202,7 +203,7 @@ function getUrl() {
...
   return url;
```
IshaniSanturkar, line 207: I know this is kind of addressed by Connor's PR but could we also highlight the nodes with the token? We could pass some version of roots again like we did before and highlight those nodes.

src/main/java/com/google/sps/DataGraph.java
```diff
@@ -352,49 +404,70 @@ private void dfsVisit(GraphNode gn, Map<GraphNode, Boolean> visited, int depthRe
...
+        continue;
+      }
+      if (graphNodesMap.containsKey(name)) {
+        GraphNode tgtNode = graphNodesMap.get(name);
+        nextLayerChildren.add(tgtNode);
+        nextLayerParents.add(tgtNode);
+      }
+    }
+    // Nothing was searched (no nodeName or tokens)
+    if (names.size() == 1 && noNodeNameSearched) {
```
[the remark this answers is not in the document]
haoyxc: that would mean that the token map has an empty string as a node right? which shouldn't happen?

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -158,16 +163,17 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-    // Get the graph at the requested mutation number and truncate it
-    try {
-      currDataGraph =
-          Utility.getGraphAtMutationNumber(
-              originalDataGraph, currDataGraph, mutationNumber, mutList);
-    } catch (IllegalArgumentException e) {
-      response.setHeader("serverError", e.getMessage());
-      return;
+    // Get the graph at the requested mutation number
+    currDataGraph =
```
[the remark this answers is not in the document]
haoyxc: Oh i think I did that when we had that function called multiple times and we only used the try catch that one time.

src/main/java/com/google/sps/ReachableNodesTest.java
```diff
@@ -297,4 +308,167 @@ public void findsShorterPath() {
     // Encapsulation
     Assert.assertEquals(6, graph.edges().size());
   }
+
+  /** An empty string as a node (that doesn't exist in the graph should be ignored) */
```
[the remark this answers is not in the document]
haoyxc: not great at grammar mostly

src/main/java/com/google/sps/DataGraph.java
haoyxc: passed in the nodes user searched for for changed the color!

src/main/webapp/script.js
haoyxc: i just change the color of those nodes! idk what's best for ui if you have other suggestions lmk lol

src/main/java/com/google/sps/DataGraph.java
```diff
@@ -295,15 +317,44 @@ private GraphNode changeNodeToken(GraphNode node, TokenMutation tokenMut) {
...
       tokenList.removeAll(tokenNames);
+      // Update the map
+      for (String tokenName : tokenNames) {
+        if (this.tokenMap().containsKey(tokenName)) {
+          Set<String> nodesWithToken = this.tokenMap().get(tokenName);
+          nodesWithToken.remove(node.name());
+          if (nodesWithToken.size() == 0) { // No more nodes with token
+            this.tokenMap().remove(tokenName);
+          } else {
+            this.tokenMap().put(tokenName, nodesWithToken);
```
EatingW, line 335: Do you need this `else`?

src/main/java/com/google/sps/DataGraph.java
```diff
@@ -295,15 +317,44 @@ private GraphNode changeNodeToken(GraphNode node, TokenMutation tokenMut) {
...
+
+      // Update the map
+      for (String tokenName : tokenNames) {
+        addToken(tokenName, node.name());
+      }
     } else if (tokenMutType == TokenMutation.Type.DELETE_TOKEN) {
       tokenList.removeAll(tokenNames);
+      // Update the map
+      for (String tokenName : tokenNames) {
+        if (this.tokenMap().containsKey(tokenName)) {
```
EatingW, line 329: I like how `addToken` is its own separate function. Maybe `removeToken` could also be a separate function?

src/main/java/com/google/sps/DataGraph.java
```diff
@@ -352,50 +403,57 @@ private void dfsVisit(GraphNode gn, Map<GraphNode, Boolean> visited, int depthRe
...
               .map(rootName -> this.graphNodesMap().get(rootName))
               .collect(Collectors.toList());
       nextLayerChildren.addAll(rootNodes);
-    } else {
-      GraphNode tgtNode = graphNodesMap.get(name);
-      nextLayerChildren.add(tgtNode);
-      nextLayerParents.add(tgtNode);
     }
 
+    for (String name : names) {
```
EatingW, line 437: Optional but you could also use Collections.stream here, with an additional filter, right?

src/main/java/com/google/sps/ReachableNodesTest.java
```diff
@@ -297,4 +308,160 @@ public void findsShorterPath() {
...
+    Set<EndpointPair<GraphNode>> graphEdges = truncatedGraph.edges();
+
+    Assert.assertEquals(0, graphNodes.size());
+
+    Assert.assertEquals(0, graphEdges.size());
+  }
+
+  /** Only consider nodes found in the graph */
+  @Test
+  public void invalidWithValidNodes() {
```
EatingW, line 419: How is this test different from `invalidNodesIgnored`?

src/main/java/com/google/sps/ReachableNodesTest.java
```diff
@@ -297,4 +308,160 @@ public void findsShorterPath() {
...
+    Assert.assertTrue(graphNodes.contains(gNodeA));
+    Assert.assertTrue(graphNodes.contains(gNodeB));
+    Assert.assertFalse(graphNodes.contains(gNodeC));
+
+    Assert.assertEquals(1, graphEdges.size());
+  }
+
+  /** When the node searched doesn't exist, empty graph is returned */
+  @Test
+  public void allNodesSearchedDontExist() {
```
EatingW, line 446: How is this different from `nonExistentNodesOnly`?

src/main/java/com/google/sps/ReachableNodesTest.java
```diff
@@ -297,4 +308,160 @@ public void findsShorterPath() {
...
+    List<String> lst = new ArrayList<>(Arrays.asList("X"));
+
+    MutableGraph<GraphNode> truncatedGraph = dataGraph.getReachableNodes(lst, 0);
+    Set<GraphNode> graphNodes = truncatedGraph.nodes();
+    Set<EndpointPair<GraphNode>> graphEdges = truncatedGraph.edges();
+
+    Assert.assertEquals(0, graphNodes.size());
+
+    Assert.assertEquals(0, graphEdges.size());
+  }
```
EatingW, line 466: I think another interesting test case to add is when you search for B, E in the following graph with radius 1, a disjoint graph of (A, B) and (D, E) is returned. 
     A
   /    \
 B      C
           \
           D
             \
             E

src/main/webapp/script.js
```diff
@@ -294,13 +297,12 @@ function getGraphDisplay(graphNodes, graphEdges, mutList, reason) {
...
 
-  // If a node is searched, color it (it's fuchsia because I thought it was pretty, but definitely open to change! )
-  const nodeFilter = document.getElementById("node-name-filter");
-  if (nodeFilter && nodeFilter.value) {
-    const target = findNodeInGraph(cy, nodeFilter.value);
-    if (target) {
-      target.style('background-color', colorScheme["filteredNodeColor"]);
-    }
+  // Color the queried nodes (it's fuchsia because I thought it was pretty, but definitely open to change! )
+  if(queriedNodes) {
```
EatingW, line 301: nit: space after `if`

src/main/webapp/script.js
```diff
@@ -294,13 +297,12 @@ function getGraphDisplay(graphNodes, graphEdges, mutList, reason) {
...
   });
 
-  // If a node is searched, color it (it's fuchsia because I thought it was pretty, but definitely open to change! )
-  const nodeFilter = document.getElementById("node-name-filter");
-  if (nodeFilter && nodeFilter.value) {
-    const target = findNodeInGraph(cy, nodeFilter.value);
-    if (target) {
-      target.style('background-color', colorScheme["filteredNodeColor"]);
-    }
+  // Color the queried nodes (it's fuchsia because I thought it was pretty, but definitely open to change! )
```
EatingW, line 300: Right now mutation display has color changes, and highlighting nodes has border-width changes. I wonder if changing both color and boarder-width for queried nodes might confuse with other forms of highlighting? What happens if a node was both queried, and has a token change mutation applied? 

I don't really have a great suggestion either so feel free to discuss among yourselves.

src/main/java/com/google/sps/ReachableNodesTest.java
haoyxc: oops I changed the function & at first there were all sorts of edge cases but I think I forgot to delete the redundant tests after. sorry!!

src/main/webapp/script.js
haoyxc: ahh yea that's definitely a great point. we'll think about it! maybe we could potentially come up with something by the time we do the task about making the frontend look nicer/improving readability?

PR #43, opened 2020-07-23
Add a timeline to navigate to different points in the list of mutations visually

IshaniSanturkar:
In this PR, I:
- added a timeline which the user can click to navigate to different indices in the relevant indices list
- added a text input where the user can input a raw mutation number and go to that graph
- integrated the above features with each other and with the next/previous buttons
- added front-end tests for the new features

Timeline:
image

Input box:
image

### DIFF
--- a/spec/jstests-spec.js
+++ b/spec/jstests-spec.js
@@ -296,12 +296,12 @@ describe("Pressing next and previous buttons associated with a graph", function 
     expect(prevButton.disabled).toBe(true);
   });
 
-  it("Correctly doesn't change anything when there aren't any mutations", function() {
+  it("Correctly doesn't change anything when there aren't any mutations", function () {
     document.body.appendChild(numDisplay);
     initializeNumMutations(-1);
     setCurrMutationNum(10);
-   
-   // Nothing changes with a negative numMutations
+
+    // Nothing changes with a negative numMutations
     navigateGraph(-1);
     expect(currMutationNum).toBe(10);
 
@@ -309,7 +309,7 @@ describe("Pressing next and previous buttons associated with a graph", function 
     expect(currMutationNum).toBe(10);
   });
 
-  it("Correctly navigates forward when index is a decimal", function() {
+  it("Correctly navigates forward when index is a decimal", function () {
     document.body.appendChild(numDisplay);
     setCurrMutationIndex(.5);
     setCurrMutationNum(0);
@@ -319,7 +319,7 @@ describe("Pressing next and previous buttons associated with a graph", function 
     expect(currMutationNum).toBe(1);
   });
 
-  it("correctly navigates backward when index is a decimal", function() {
+  it("correctly navigates backward when index is a decimal", function () {
     document.body.appendChild(numDisplay);
     setCurrMutationIndex(.5);
     setCurrMutationNum(0);
@@ -828,9 +828,8 @@ describe("Testing graph input functionality", function () {
   let totalMutNumberText;
 
   beforeEach(function () {
-    document.body.innerHTML = 
-    `
-      <div id="num-mutation-display">
+    document.body.innerHTML =
+      `<div id="num-mutation-display">
         <p id="graph-number-text">Graph </p>
         <input type="number" id="graph-number"  min="0" max="0" value="0">
         <p id="total-mutation-number-text">out of 0</p>
@@ -843,12 +842,12 @@ describe("Testing graph input functionality", function () {
 
   it("correctly updates graph number input when mutation list is empty", function () {
     setMutationIndexList([]);
-    initializeNumMutations(0);
     setCurrMutationNum(-1);
     setCurrMutationIndex(-1);
     setMaxNumMutations(0);
 
     updateGraphNumInput();
+    // Just the initial graph
     expect(graphNumberInput.value).toBe("0");
     expect(graphNumberInput.min).toBe("0");
     expect(graphNumberInput.max).toBe("0");
@@ -862,12 +861,12 @@ describe("Testing graph input functionality", function () {
 
   it("correctly updates graph number input when the mutation index list is non-empty", function () {
     setMutationIndexList([1, 5, 7, 11]);
-    initializeNumMutations(4);
     setCurrMutationNum(1);
     setCurrMutationIndex(0);
     setMaxNumMutations(20);
 
     updateGraphNumInput();
+    // Current graph is the graph after mutation 1
     expect(graphNumberInput.value).toBe("2");
     expect(graphNumberInput.min).toBe("0");
     expect(graphNumberInput.max).toBe("20");
--- a/spec/jstests-spec.js
+++ b/spec/jstests-spec.js
@@ -1,9 +1,11 @@
+
 import {
   initializeNumMutations, setMutationIndexList, setCurrMutationNum, setCurrMutationIndex,
   initializeTippy, generateGraph, getUrl, navigateGraph, currMutationNum, currMutationIndex,
-  numMutations, updateButtons, searchNode, highlightDiff, initializeReasonTooltip, getGraphDisplay,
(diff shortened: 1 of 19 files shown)


spec/jstests-spec.js
```diff
@@ -672,4 +673,189 @@ describe("Checking binary search functions", function () {
...
+});
+
+describe("Testing slider functionality", function () {
+  beforeAll(function () {
+    document.body.innerHTML = '';
+  });
+
+  beforeEach(function () {
+    document.body.innerHTML = `
+    <div id="slider" class="mdc-slider mdc-slider--discrete mdc-slider--display-markers" tabindex="0" role="slider" aria-valuemin="0" aria-valuemax="0" aria-valuenow="0" aria-label="Select Value">
```
haoyxc, line 685: do you think it could be helpful to import this string as a constant?

spec/jstests-spec.js
```diff
@@ -672,4 +673,189 @@ describe("Checking binary search functions", function () {
...
+    const prevButton = document.createElement("button");
+    prevButton.id = "prevbutton";
+    prevButton.onclick = () => { navigateGraph(-1); updateButtons(); };
+    const nextButton = document.createElement("button");
+    nextButton.id = "nextbutton";
+    nextButton.onclick = () => { navigateGraph(1); updateButtons(); };
+    document.body.appendChild(prevButton);
+    document.body.appendChild(nextButton);
+
+
```
haoyxc, line 767: nit: extra line

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -187,7 +187,7 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     }
     // The searched node exists but is not mutated in the current graph
     if (truncatedGraph.nodes().size() != 0
-        && mutationNumber != -1
+        && !(mutationNumber == -1 && nodeNameParam.equals(""))
```
haoyxc, line 190: do you think it could make sense to separate this into 2 separate statements? I don't think the state of mutationNumber and nodeNameParam are necessarily linked?

src/main/webapp/index.html
```diff
@@ -36,19 +45,36 @@ <h1>
...
-      <label>Search by Node: </label>
-      <input type="text" id="search" name="search"></input>
-      <button id="search-button">Search</button>
-      <label id="search-error"></label>
     </div>
     <!-- COLUMN 2: includes messages -->
     <div class="graph-content-column" id="graph-content-logs">
       <ul id="log-list">
-        <li class="log-msg">Welcome to our application! Here you'll see a log of messages pertaining to the graph on the left. Enjoy! :)</li>
+        <li>Welcome to our application! Here you'll see a log of messages pertaining to the graph on the left. Enjoy! :)
```
haoyxc, line 74: do you think we would ever make use of the class?

src/main/webapp/script.js
```diff
@@ -55,9 +58,25 @@ let numMutations = 0;
...
-let totalNum = 0;
+/**
+ * Initializes the mutation index slider upon document load and sets it up to
+ * regenerate the graph when its value is changed
+ */
+function initializeSlider() {
+  mutationNumSlider = new MDCSlider(document.querySelector('.mdc-slider'));
+  mutationNumSlider.listen('MDCSlider:change', () => {
+    currMutationIndex = mutationNumSlider.value;
+    currMutationNum = (currMutationIndex === -1) ? -1 : mutationIndexList[currMutationIndex];
```
haoyxc, line 76: do you need to make sure currMutationIndex < mutationIndexList.length (i mean I feel like that should that never happen but maybe as a precautionary measure haha)

src/main/webapp/script.js
```diff
@@ -130,6 +158,7 @@ async function generateGraph() {
 
   mutationIndexList = JSON.parse(jsonResponse.mutationIndices);
   numMutations = mutationIndexList.length;
+  maxNumMutations = jsonResponse.totalMutNumber;
```
haoyxc, line 161: is there any reason you created a setter but didn't use it?

src/main/webapp/script.js
```diff
@@ -797,3 +849,46 @@ function getClosestIndices(indicesList, element) {
...
+  if (!graphNumberInput || graphNumberInput.length === 0) {
+    return;
+  }
+  if (graphNumberInput.value >= maxNumMutations) {
+    graphNumberInput.value = maxNumMutations;
+  }
+  if (graphNumberInput.value <= 0) {
+    graphNumberInput.value = 0;
+  }
+  currMutationNum = graphNumberInput.value - 1;
```
haoyxc, line 893: do you think it could be easier to get rid of this one off? and have the value they put in just be the currMutationNum?

src/main/webapp/script.js
```diff
@@ -96,11 +122,14 @@ async function generateGraph() {
   let graphNodes = [];
   let graphEdges = [];
 
-  // disable buttons
+  // disable all possible input fields
```
EatingW, line 125: nit: I feel like this section is getting large enough to deserve its own method.

src/main/webapp/script.js
```diff
@@ -333,7 +386,7 @@ function getGraphDisplay(graphNodes, graphEdges, mutList, reason) {
   // Zoom in on them and activate their reason tooltips
   makeInteractiveAndFocus(cy, elems);
 
-  
+
```
EatingW, line 389: nit: delete?

spec/jstests-spec.js
IshaniSanturkar: I just copy pasted this from the material design slider reference. Do you think it would be weird to put it in script given that it doesn't have much to do with our code?

src/main/java/com/google/sps/DataServlet.java
IshaniSanturkar: If mutationNumber is -1 and nodeNameParam is not empty then this case should always happen because a real node is never mutated in the initial graph. The only reason I had that check is to deal with the case when we are at the initial graph but we are not searching for a node. In that case no error message should be given. How do you think we should separate it?

src/main/webapp/index.html
IshaniSanturkar: Oops I think it got deleted by accident while merging. Sorry!

src/main/webapp/script.js
IshaniSanturkar: I used it in the test file. I don't know if it makes a lot of sense to use a setter in the same file as the variable. Do you think I should do that?

src/main/webapp/script.js
IshaniSanturkar: I just did it because this input box is part of the "Graph X of XX" display and I thought it would be more intuitive for a user to say "Graph 1" than "Graph 0".

spec/jstests-spec.js
```diff
@@ -1,8 +1,9 @@
 import {
-  initializeNumMutations, setMutationIndexList, setCurrMutationNum, setCurrMutationIndex, initializeTippy,
-  generateGraph, getUrl, navigateGraph, currMutationNum, currMutationIndex, numMutations,
-  updateButtons, searchNode, highlightDiff, initializeReasonTooltip, getGraphDisplay,
-  getClosestIndices
+  initializeNumMutations, setMutationIndexList, setCurrMutationNum, setCurrMutationIndex,
```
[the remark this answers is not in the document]
IshaniSanturkar: We'll refactor this in the readability PR :)

spec/jstests-spec.js
haoyxc: Haha you're probably right it's fine!

src/main/java/com/google/sps/DataServlet.java
haoyxc: Or I guess it would be something like `mutationNumber != -1 || !nodeNameParam.equals(""))` if I'm distributing correctly. But i think we stuck with this in later branches so it's fine to keep!

PR #45, opened 2020-07-24
Clicking next on a filtered/depth-limited graph goes to next graph where any displayed node is mutated

IshaniSanturkar:
In this PR, I:
- added code so that when the user filters a graph by node name or limits the radius of display, clicking next goes to the next graph where any on-screen nodes are mutated instead of going to the next graph where any node (possibly not displayed) is mutated.
- added backend tests for the added logic

### DIFF
--- a/src/main/java/com/google/sps/DataServlet.java
+++ b/src/main/java/com/google/sps/DataServlet.java
@@ -232,8 +232,8 @@ public class DataServlet extends HttpServlet {
         && diff.getMutationList().size() == 0) {
       response.setHeader(
           "serverMessage",
-          "The desired set of nodes is mutated in this graph but your other parameters (for"
-              + " eg. radius), limit the display of the mutations. Please try increasing your radius"
+          "The desired set of nodes is mutated in this graph but your other parameters (for eg."
+              + " radius), limit the display of the mutations. Please try increasing your radius"
               + " to view the mutation.");
     }
 
--- a/src/main/java/com/google/sps/DataServlet.java
+++ b/src/main/java/com/google/sps/DataServlet.java
@@ -71,12 +71,9 @@ public class DataServlet extends HttpServlet {
      *********************************
      */
     if (currDataGraph == null && originalDataGraph == null) {
-      success =
-          initializeGraphVariables(
-              getServletContext().getResourceAsStream("/WEB-INF/graph.textproto"));
+      success = initializeGraphVariables(getServletContext().getResourceAsStream("/WEB-INF/graph.textproto"));
       if (!success) {
-        response.setHeader(
-            "serverError", "Failed to parse input graph into Guava graph - not a DAG!");
+        response.setHeader("serverError", "Failed to parse input graph into Guava graph - not a DAG!");
         return;
       }
       currDataGraph = originalDataGraph.getCopy();
@@ -91,10 +88,9 @@ public class DataServlet extends HttpServlet {
      *************************************
      */
     if (mutList == null) {
-      initializeMutationVariables(
-          getServletContext().getResourceAsStream("/WEB-INF/mutation.textproto"));
+      initializeMutationVariables(getServletContext().getResourceAsStream("/WEB-INF/mutation.textproto"));
       // Populate the list of all possible mutation indices
-      defaultIndices = IntStream.range(0, mutList.size() - 1).boxed().collect(Collectors.toList());
+      defaultIndices = IntStream.range(0, mutList.size()).boxed().collect(Collectors.toList());
       // and store this as the list of relevant indices for filtering by empty string
       // (= not filtering)
       mutationIndicesMap.put("", defaultIndices);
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/sps/NodeMutationFilterTest.java
```diff
@@ -88,4 +92,247 @@ public void getMutationsOfNull() {
...
+
+    List<Integer> truncatedList =
+        Utility.findRelevantMutations(nodeNames, mutationIndicesMap, multiMutList);
+
+    Assert.assertEquals(4, truncatedList.size());
+    Assert.assertTrue(truncatedList.get(0) == 0);
+    Assert.assertTrue(truncatedList.get(1) == 1);
+    Assert.assertTrue(truncatedList.get(2) == 2);
+    Assert.assertTrue(truncatedList.get(3) == 3);
+  }
```
haoyxc, line 337: could add tests for empty nodeNames as well! Or empty multiMutList. Or when none of the nodeNames are found in the mutationIndicesMap.

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -169,16 +161,38 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
           "serverError", "The searched node does not exist anywhere in this graph or in mutations");
       return;
     }
     // The searched node is not in the graph but is mutated at some past/future
-    // point
-    if (truncatedGraph.nodes().size() == 0 && filteredMutationIndices.size() != 0) {
+    // point. The diff conditions are included to prevent entry into this case
+    // when the searched node is deleted for example. The diff being non-empty
+    // means that there is some mutation pertaining to the searched node to show
+    if (truncatedGraph.nodes().size() == 0
```
haoyxc, line 193: Is there any way to make this expression account for deleted nodes?

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -169,16 +161,38 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     }
     truncatedGraph = currDataGraph.getReachableNodes(nodeNameParam, depthNumber);
 
-    // We set the headers in the following 3 scenarios:
+    if (nodeNameParam.equals("") && truncatedGraph.equals(currDataGraph.graph())) {
```
haoyxc, line 164: sometimes you use nodeNameParam.length() == 0 and other times .equals(""). Maybe just stick to one or the other for consistency?

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -169,16 +161,38 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+      // If we are not filtering the graph or limiting its depth, show all mutations of all nodes
+      filteredMutationIndices = defaultIndices;
+    } else {
+      // Get the names of all the displayed nodes and find all indices of mutations
+      // that mutate any of them
+      Set<String> truncatedGraphNodeNames = Utility.getNodeNamesInGraph(truncatedGraph);
+      // Also get mutations relevant to the searched node if it is not an empty string
+      if (nodeNameParam.length() != 0) {
+        truncatedGraphNodeNames.add(nodeNameParam);
```
haoyxc, line 173: if nodeNameParam is not empty (something is searched),  shouldn't it already be in the truncated graph? (since nodeNameParam was passed into the collection that the reachable nodes function took)

src/main/java/com/google/sps/Utility.java
```diff
@@ -232,4 +235,37 @@ public static MultiMutation filterMultiMutationByNodes(MultiMutation mm, Set<Str
         .setReason(mm.getReason())
         .build();
   }
+
+  /**
+   * Given a list of node names, a map from node name to mutation indices of that node and a list of
+   * multimutations applied to all nodes, returns a list of indices of multimutations in which any
+   * of the nodes in nodeNames get mutated (returned in sorted order)
+   *
+   * @param nodeNames the names of nodes to restrict the returned list of mutations to
```
haoyxc, line 244: or if you reeeaaaaally wanted to include the "", consider adding that into the comment?

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -73,7 +73,7 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     if (currDataGraph == null && originalDataGraph == null) {
       success =
           initializeGraphVariables(
-              getServletContext().getResourceAsStream("/WEB-INF/initial_graph.textproto"));
+              getServletContext().getResourceAsStream("/WEB-INF/graph.textproto"));
```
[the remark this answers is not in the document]
IshaniSanturkar: It was just easier to test on the smaller graph. I'll change it back, sorry!

src/main/java/com/google/sps/NodeMutationFilterTest.java
IshaniSanturkar: What do you think should happen when nodeNames is empty?

src/main/java/com/google/sps/DataServlet.java
IshaniSanturkar: Oops I think I removed the extra diff check mentioned in the comment thinking that it was redundant.

src/main/java/com/google/sps/DataServlet.java
IshaniSanturkar: Good point :)

src/main/java/com/google/sps/DataServlet.java
IshaniSanturkar: I think this was just for highlighting the diff in case the searched node was deleted. Open to better ways of doing this :)

src/main/java/com/google/sps/Utility.java
IshaniSanturkar: Haha no I think you're right

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -187,19 +202,23 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-      diff = Utility.filterMultiMutationByNodes(diff, truncatedGraphNodeNames);
+    // There is a diff between the previously-displayed graph and the current graph but
+    // no mutations in it only mutate on-screen nodes
+    if (filteredMutationIndices.indexOf(mutationNumber) != -1
+        && diff != null
+        && diff.getMutationList().size() == 0) {
+      response.setHeader(
+          "serverMessage",
+          "The desired set of nodes is mutated in this graph but your other parameters (for"
+              + " eg.depth), limit the display of the mutations. Please try increasing your radius"
```
EatingW, line 220: nit: depth or radius?

src/main/java/com/google/sps/NodeMutationFilterTest.java
```diff
@@ -88,4 +92,339 @@ public void getMutationsOfNull() {
...
+    MultiMutation addHM = MultiMutation.newBuilder().addMutation(addH).addMutation(addHG).build();
+    MultiMutation addDGM = MultiMutation.newBuilder().addMutation(addDG).build();
+    MultiMutation addTokenToBM = MultiMutation.newBuilder().addMutation(addTokenToB).build();
+
+    List<MultiMutation> multiMutList = new ArrayList<>();
+    multiMutList.add(removeFM);
+    multiMutList.add(addGM);
+    multiMutList.add(addHM);
+    multiMutList.add(addDGM);
+    multiMutList.add(addTokenToBM);
```
EatingW, line 160: Seems like this set up for the multi mutation list is common to a couple of your tests. You can create a private helper function for this that returns a multi mutation list. Otherwise the tests get very long and redundant.

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -169,16 +161,39 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    } else {
+      // Get the names of all the displayed nodes and find all indices of mutations
+      // that mutate any of them
+      Set<String> truncatedGraphNodeNames = Utility.getNodeNamesInGraph(truncatedGraph);
+      // Also get mutations relevant to the searched node if it is not an empty string
+      if (nodeNameParam.length() != 0) {
+        truncatedGraphNodeNames.add(nodeNameParam);
+      }
+      filteredMutationIndices =
+          Utility.findRelevantMutations(truncatedGraphNodeNames, mutationIndicesMap, mutList);
```
EatingW, line 176: similar to what we discussed during the team meeting today, here the `truncatedGraphNodeNames` should be the ones without lines 172-174. If the `nodeNameParam` is deleted, you only want to show the indices relevant to the rest of the graph. 
The one on line 179 should be the nodeNames including `nodeNameParam`. Seems like in this case swapping this line with 172-174 would work, but also feel free to integrate this with Cindy's in a later PR.

src/main/java/com/google/sps/NodeMutationFilterTest.java
haoyxc: either all or nothing, but I think all makes more sense because nothing could also mean that none of the nodeNames are contained. What do you think?

src/main/java/com/google/sps/DataServlet.java
haoyxc: I think this is what we worked off of for other branches so it seems fine!!

src/main/java/com/google/sps/NodeMutationFilterTest.java
IshaniSanturkar: All makes sense!

IshaniSanturkar: Actually nothing makes more sense. For example in the case where the node we search for gets deleted in this graph, the truncated graph nodes is empty since the deleted node is used as a root. So finding all the relevant mutations would give us all indices, which is slightly unintuitive.

src/main/java/com/google/sps/DataServlet.java
IshaniSanturkar: Since it's not that big a change I made it here :)

IshaniSanturkar: Actually, what if the searched node doesn't exist yet but is added in the future. Then if we changed the logic we would never be able to detect it.

src/main/java/com/google/sps/NodeMutationFilterTest.java
IshaniSanturkar: Never mind, I'm doing all :)

src/main/java/com/google/sps/DataServlet.java
IshaniSanturkar: I think @haoyxc addressed this in her PR, so resolving this for now!

PR #47, opened 2020-07-28
Allow filtering by multiple nodes (through filtering by tokens or comma separated)

haoyxc:
Changes in this PR 
- Integrate next/prev buttons & timeline with filtering by token across multiple graphs
- Include function that gets the mutations for a token, and gets the mutations of a collection of nodes
- Add tests for new functionality
- Modify headers to be more clear
- Filter the diff by truncatedNodeNames and the nodes with the token
- Allow user to input multiple nodes to filter by separated by commas
- Change singular nodeName to nodeNames
- Pass in stringified array
- Change frontend to reflect functionality
- Update/add frontend tests
- Fix integration between token / node integration in DataServlet 

Filtering by JZBOxLeUAkj5cdKDoYRFbULqWRZYWgeqaqAHyIIslaLJE5dyQr4vMrp1u85SS8HcxMHDsyyCBg
[capture.zip](https://github.com/googleinterns/step49-2020/files/4989931/capture.zip)
Screenshot 2020-07-28 at 12 06 39 PM
Screenshot 2020-07-28 at 12 06 46 PM

If filtering by "A, B"
Screenshot 2020-07-28 at 5 51 06 PM
[capture (1).zip](https://github.com/googleinterns/step49-2020/files/4991564/capture.1.zip)

### DIFF
--- a/spec/jstests-spec.js
+++ b/spec/jstests-spec.js
@@ -404,6 +404,17 @@ describe("Check correct url params", function () {
     expect(constructedUrl.get("nodeNames")).toEqual('["Annie boo","B"]');
   });
 
+  it ("does not pass empty names", function () {
+    nodeName.value = "A, , ,,";
+    document.body.appendChild(nodeName); 
+    const requestString = getUrl(); 
+    const requestParams = requestString.substring(requestString.indexOf("?"));
+
+    const constructedUrl = new URLSearchParams(requestParams);
+    expect(constructedUrl.has("nodeNames")).toBe(true); 
+    expect(constructedUrl.get("nodeNames")).toEqual('["A"]');
+  })
+
   it("passes correct tokenName when tokenName has a value", function () {
     tokenName.value = "1";
     document.body.appendChild(tokenName);
--- a/spec/jstests-spec.js
+++ b/spec/jstests-spec.js
@@ -395,8 +395,8 @@ describe("Check correct url params", function () {
     expect(constructedUrl.get("mutationNum")).toBe("1");
 
     // Not on page here, should be empty
-    expect(constructedUrl.has("nodeName")).toBe(true);
-    expect(constructedUrl.get("nodeName")).toBe("");
+    expect(constructedUrl.has("nodeNames")).toBe(true);
+    expect(constructedUrl.get("nodeNames")).toEqual('[""]');
 
     expect(constructedUrl.has("tokenName")).toBe(true);
     expect(constructedUrl.get("tokenName")).toBe("");
@@ -404,20 +404,44 @@ describe("Check correct url params", function () {
 
   it("passes correct nodeName when nodeName has a value", function () {
     nodeName.value = "A";
-    document.body.appendChild(nodeName);
+    document.body.appendChild(nodeName); 
 
-    const requestString = getUrl();
+    const requestString = getUrl(); 
     const requestParams = requestString.substring(requestString.indexOf("?"));
 
     const constructedUrl = new URLSearchParams(requestParams);
-    expect(constructedUrl.has("nodeName")).toBe(true);
-    expect(constructedUrl.get("nodeName")).toBe("A");
+    expect(constructedUrl.has("nodeNames")).toBe(true); 
+    expect(constructedUrl.get("nodeNames")).toEqual('["A"]');
+  });
+
+  it("passes correct nodeName when nodeName has a comma separated value", function () {
+    nodeName.value = "A,B";
+    document.body.appendChild(nodeName); 
+
+    const requestString = getUrl(); 
+    const requestParams = requestString.substring(requestString.indexOf("?"));
+
+    const constructedUrl = new URLSearchParams(requestParams);
+    expect(constructedUrl.has("nodeNames")).toBe(true); 
+    expect(constructedUrl.get("nodeNames")).toEqual('["A","B"]');
+  });
+
+    it("passes correct nodeName when nodeName has a comma separated value and spaces", function () {
+    nodeName.value = "  Annie boo  , B";
+    document.body.appendChild(nodeName);  
+
+    const requestString = getUrl(); 
+    const requestParams = requestString.substring(requestString.indexOf("?"));
+
+    const constructedUrl = new URLSearchParams(requestParams);
+    expect(constructedUrl.has("nodeNames")).toBe(true); 
+    expect(constructedUrl.get("nodeNames")).toEqual('["Annie boo","B"]');
   });
 
   it("passes correct tokenName when tokenName has a value", function () {
     tokenName.value = "1";
     document.body.appendChild(tokenName);
-
+ 
     const requestString = getUrl();
     const requestParams = requestString.substring(requestString.indexOf("?"));
     const constructedUrl = new URLSearchParams(requestParams);
--- a/src/main/java/com/google/sps/DataServlet.java
+++ b/src/main/java/com/google/sps/DataServlet.java
@@ -74,7 +74,7 @@ public class DataServlet extends HttpServlet {
     if (currDataGraph == null && originalDataGraph == null) {
       success =
           initializeGraphVariables(
-              getServletContext().getResourceAsStream("/WEB-INF/graph.textproto"));
+              getServletContext().getResourceAsStream("/WEB-INF/initial_graph.textproto"));
       if (!success) {
         response.setHeader(
             "serverError", "Failed to parse input graph into Guava graph - not a DAG!");
@@ -93,7 +93,7 @@ public class DataServlet extends HttpServlet {
      */
     if (mutList == null) {
       initializeMutationVariables(
-          getServletContext().getResourceAsStream("/WEB-INF/mutation.textproto"));
+          getServletContext().getResourceAsStream("/WEB-INF/mutations.textproto"));
       // Populate the list of all possible mutation indices
       defaultIndices = IntStream.range(0, mutList.size()).boxed().collect(Collectors.toList());
       // and store this as the list of relevant indices for filtering by empty string
--- a/src/main/java/com/google/sps/DataServlet.java
+++ b/src/main/java/com/google/sps/DataServlet.java
@@ -14,6 +14,9 @@
 
 package com.google.sps;
 
+import com.google.gson.JsonParser;
+import com.google.gson.JsonArray;
+
 import com.google.common.collect.Sets;
 
 import java.io.IOException;
@@ -77,7 +80,7 @@ public class DataServlet extends HttpServlet {
     if (currDataGraph == null && originalDataGraph == null) {
       success =
           initializeGraphVariables(
-              getServletContext().getResourceAsStream("/WEB-INF/initial_graph.textproto"));
+              getServletContext().getResourceAsStream("/WEB-INF/graph.textproto"));
       if (!success) {
         response.setHeader(
             "serverError", "Failed to parse input graph into Guava graph - not a DAG!");
@@ -96,7 +99,7 @@ public class DataServlet extends HttpServlet {
      */
     if (mutList == null) {
       initializeMutationVariables(
-          getServletContext().getResourceAsStream("/WEB-INF/mutations.textproto"));
+          getServletContext().getResourceAsStream("/WEB-INF/mutation.textproto"));
       // Populate the list of all possible mutation indices
       defaultIndices = IntStream.range(0, mutList.size()).boxed().collect(Collectors.toList());
       // and store this as the list of relevant indices for filtering by empty string
@@ -158,12 +161,25 @@ public class DataServlet extends HttpServlet {
     if (mutationNumber > currDataGraph.numMutations()) {
       diff = Utility.getMultiMutationAtIndex(mutList, mutationNumber);
     }
+    List<String> nodeNames = new ArrayList<>();
+    try {
+      JsonParser jsonParser = new JsonParser();
+      JsonArray nodeNameArr = jsonParser.parse(nodeNameParam).getAsJsonArray();
+      for (int i = 0; i < nodeNameArr.size(); i++) {
+        String curr = nodeNameArr.get(i).getAsString().trim();
+        if (curr.length() > 0) {
+          nodeNames.add(curr);
+        }
+      }
+    } catch (IllegalStateException e) {
+    }
+    System.out.println(nodeNames);
 
     // A list of "roots" to return nodes at most depth radius from
     HashSet<String> queried = new HashSet<>();
     // We start by adding the node name if it was searched for
     if (nodeNameParam.length() > 0) {
-      queried.add(nodeNameParam);
+      queried.addAll(nodeNames);
     }
 
     // Show mutations relevant to nodes that contain the token in the current graph
@@ -200,7 +216,7 @@ public class DataServlet extends HttpServlet {
       Set<String> truncatedGraphNodeNames = Utility.getNodeNamesInGraph(truncatedGraph);
       // Also get mutations relevant to the searched node if it is not an empty string
       if (nodeNameParam.length() != 0) {
-        truncatedGraphNodeNames.add(nodeNameParam);
+        truncatedGraphNodeNames.addAll(nodeNames);
       }
 
       // A set containing a indices where nodes currently displayed on the graph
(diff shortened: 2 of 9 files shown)


spec/jstests-spec.js
```diff
@@ -361,29 +361,53 @@ describe("Check correct url params", function () {
...
 
     const constructedUrl = new URLSearchParams(requestParams);
-    expect(constructedUrl.has("nodeName")).toBe(true);
-    expect(constructedUrl.get("nodeName")).toBe("A");
+    expect(constructedUrl.has("nodeNames")).toBe(true); 
+    expect(constructedUrl.get("nodeNames")).toEqual('["A","B"]');
+  });
+
+    it("passes correct nodeName when nodeName has a comma separated value and spaces", function () {
+    nodeName.value = "  Annie boo  , B";
```
IshaniSanturkar, line 396: Interesting file name choice :)

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -77,7 +81,7 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     if (currDataGraph == null && originalDataGraph == null) {
       success =
           initializeGraphVariables(
-              getServletContext().getResourceAsStream("/WEB-INF/initial_graph.textproto"));
+              getServletContext().getResourceAsStream("/WEB-INF/graph.textproto"));
```
IshaniSanturkar, line 84: Change it back to big graph?

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -158,12 +162,28 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     if (mutationNumber > currDataGraph.numMutations()) {
       diff = Utility.getMultiMutationAtIndex(mutList, mutationNumber);
     }
+    List<String> nodeNames = new ArrayList<>();
+    try {
+      JsonParser jsonParser = new JsonParser();
+      JsonArray nodeNameArr = jsonParser.parse(nodeNamesParam).getAsJsonArray();
+      for (int i = 0; i < nodeNameArr.size(); i++) {
+        String curr = nodeNameArr.get(i).getAsString().trim();
```
IshaniSanturkar, line 170: Is the trim here pointless given that you trim on the front-end?

src/main/webapp/index.html
```diff
@@ -16,8 +16,8 @@ <h1>
   <div id="search-box">
     <label for="layers">Radius from filtered node(s) (0-20, default is 3):</label>
     <input type="number" id="num-layers" name="layers" min="0" max="20" value="3">
-    <label for="search-name">Name of node to filter for</label>
-    <input type="text" id="node-name-filter" name="search-name" placeholder="Node Name"></input>
+    <label for="search-name">Name of node(s) to filter for (enter multiple separated by commas if you would like!) </label>
```
IshaniSanturkar, line 19: I think the language should be more strict about them using commas?

spec/jstests-spec.js
haoyxc: idk why this was the first thing that came to mind!!!

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -153,8 +162,36 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     if (mutationNumber > currDataGraph.numMutations()) {
       diff = Utility.getMultiMutationAtIndex(mutList, mutationNumber);
     }
+    List<String> nodeNames = new ArrayList<>();
+    try {
+      JsonParser jsonParser = new JsonParser();
```
IshaniSanturkar, line 167: I realized that when I was writing this I was using older documentation and this constructor is now deprecated. Could you replace these two lines with a call to [`JsonParser.parseString()`](https://javadoc.io/doc/com.google.code.gson/gson/latest/com.google.gson/com/google/gson/JsonParser.html#parseString(java.lang.String)). The new method also throws a different exception so we'll have to check for that too.

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -164,37 +201,59 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+      // or queried are mutated
+      Set<Integer> mutationIndicesSet = new HashSet<>();
+
+      mutationIndicesSet.addAll(
+          Utility.findRelevantMutations(truncatedGraphNodeNamesNext, mutationIndicesMap, mutList));
+      mutationIndicesSet.addAll(Utility.getMutationIndicesOfToken(tokenParam, mutList));
+      filteredMutationIndices = new ArrayList<>(mutationIndicesSet);
+      Collections.sort(filteredMutationIndices);
+
+      if (tokenParam.length() == 0) {
```
IshaniSanturkar, line 248: Can we eliminate the check on line 234 by just filtering by the union of the original `truncatedGraphNodeNames` and `queried`? In the case that no token is searched, queried will just contain the node names if that list is non-empty?

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -164,37 +201,59 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
     truncatedGraph = currDataGraph.getReachableNodes(queried, depthNumber);
 
-    if (nodeNameParam.length() == 0 && truncatedGraph.equals(currDataGraph.graph())) {
-      // If we are not filtering the graph or limiting its depth, show all mutations of all nodes
+    // To get the nodes to calculate relevant mutations from. If queried and queried next contain
+    // the same
+    // nodes, then no reason to regenerate the graph
+    MutableGraph<GraphNode> truncatedGraphNext =
+        queried.equals(queriedNext)
+            ? Graphs.copyOf(truncatedGraph)
```
IshaniSanturkar, line 218: Is there a reason why we create a copy when queried is equal to queried next. I don't think we mutate either so there shouldn't be a need to copy the internal data structures?

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -212,33 +271,36 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
           "serverMessage",
           "The desired set of nodes is mutated in this graph but your other parameters (for eg."
               + " radius), limit the display of the mutations. Please try increasing your radius"
               + " to view the mutation.");
     }
-
     graphJson =
-        Utility.graphToJson(truncatedGraph, filteredMutationIndices, diff, mutList.size(), queried);
+        Utility.graphToJson(
+            truncatedGraph, filteredMutationIndices, filteredDiff, mutList.size(), queried);
```
IshaniSanturkar, line 303: Should we return queried or queriedNext?

src/main/java/com/google/sps/Utility.java
```diff
@@ -64,7 +64,7 @@ public static GraphNode protoNodeToGraphNode(Node thisNode) {
    * @param mutationIndices the indices in the entire mutation list that mutate the relevant nodes
    * @param mutDiff the difference between the current graph and the requested graph
    * @param maxNumber the total number of mutations, without filtering
-   * @param queried a list of node names the client had requested
+   * @param queried a set of node names the client had requested
    * @return a JSON object containing as entries the nodes and edges of this graph as well as the
    *     length of the list of mutations this graph is an intermediate result of applying, the
    *     indices at which relevant nodes are mutated and the change made to relevant nodes to obtain
```
IshaniSanturkar, line 70: Also document the new fields we now return here.

src/main/webapp/WEB-INF/mutation.textproto
```diff
@@ -67,5 +80,22 @@ mutation {
...
+  }
+}
+reason: "Changing tokens of node B"
+}
+mutation {
+mutation {
+  type: DELETE_NODE
+  start_node: "I"
+}
+reason: "Add node I"
```
IshaniSanturkar, line 100: Delete node I.

src/main/java/com/google/sps/DataServlet.java
haoyxc: I think it is queriedNext, right?

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -153,8 +161,35 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    List<String> nodeNames = new ArrayList<>();
+    try {
+      JsonArray nodeNameArr = JsonParser.parseString(nodeNamesParam).getAsJsonArray();
+      for (int i = 0; i < nodeNameArr.size(); i++) {
+        String curr = nodeNameArr.get(i).getAsString().trim();
+        if (curr.length() > 0) {
+          nodeNames.add(curr);
+        }
+      }
+    } catch (JsonSyntaxException e) {
```
IshaniSanturkar, line 173: I think the `getAsJsonArray` also throws an IllegalStateException which we used to handle?

src/main/java/com/google/sps/Utility.java
```diff
@@ -249,22 +276,22 @@ public static MultiMutation filterMultiMutationByNodes(MultiMutation mm, Set<Str
 
   /**
    * Given a list of node names, a map from node name to mutation indices of that node and a list of
-   * multimutations applied to all nodes, returns a list of indices of multimutations in which any
-   * of the nodes in nodeNames get mutated (returned in sorted order)
+   * multimutations applied to all nodes, returns a set of indices of multimutations in which any of
+   * the nodes in nodeNames get mutated. if nodeNames is empty, an empty set is returned
```
IshaniSanturkar, line 280: Nit: capital I in if

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -153,8 +162,35 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    List<String> nodeNames = new ArrayList<>();
+    try {
+      JsonArray nodeNameArr = JsonParser.parseString(nodeNamesParam).getAsJsonArray();
+      for (int i = 0; i < nodeNameArr.size(); i++) {
+        String curr = nodeNameArr.get(i).getAsString().trim();
+        if (curr.length() > 0) {
+          nodeNames.add(curr);
+        }
+      }
+    } catch (JsonSyntaxException | IllegalStateException e) {
```
EatingW, line 174: Does this mean the input is invalid? If so, you should probably propagate the error to the user.

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -153,8 +162,35 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+
+    // A list of "roots" to return nodes at most depth radius from
+    HashSet<String> queried = new HashSet<>();
+
+    // roots to calculate the mutations from
+    HashSet<String> queriedNext = new HashSet<>();
+    // We start by adding the node name if it was searched for
+    if (nodeNames.size() > 0) {
+      queried.addAll(nodeNames);
+      queriedNext.addAll(nodeNames);
```
EatingW, line 185: What if the some of the nodeNames were deleted in this mutation? Should queriedNext still contain these nodes?

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -164,37 +200,57 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
       queried.addAll(currDataGraph.tokenMap().get(tokenParam));
+      queriedNext.addAll(currDataGraph.tokenMap().get(tokenParam));
     }
+
+    // Truncate the graph from the nodes that the client had searched for
     truncatedGraph = currDataGraph.getReachableNodes(queried, depthNumber);
 
-    if (nodeNameParam.length() == 0 && truncatedGraph.equals(currDataGraph.graph())) {
-      // If we are not filtering the graph or limiting its depth, show all mutations of all nodes
+    // The nodes to calculate relevant mutations from
```
EatingW, line 212: Since this is a `truncatedGraph`, this comment is a bit inaccurate since it seems to indicate "a list of nodes" :)

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -212,33 +268,36 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
         && (diff == null || diff.getMutationList().size() == 0)) {
```
EatingW, line 268: Shouldn't this be filteredDiff now?

src/main/java/com/google/sps/Utility.java
```diff
@@ -207,6 +205,35 @@ public static MultiMutation getMultiMutationAtIndex(List<MultiMutation> multiMut
...
+  /**
+   * Returns a set of indices on the original list that related to a given token
+   *
+   * @param tokenName the token name to search for
+   * @param origList the original list of mutations
+   * @return a set of indices, empty if tokenName is null or if token is not changed
+   */
+  public static Set<Integer> getMutationIndicesOfToken(
+      String tokenName, List<MultiMutation> origList) {
+    Set<Integer> lst = new HashSet<>();
```
EatingW, line 217: nit: `lst` doesn't seem like a representative name for a set :)

src/main/java/com/google/sps/Utility.java
```diff
@@ -207,6 +205,35 @@ public static MultiMutation getMultiMutationAtIndex(List<MultiMutation> multiMut
...
+    if (tokenName == null || tokenName.length() == 0) {
+      return lst;
+    }
+    for (int i = 0; i < origList.size(); i++) {
+      MultiMutation multiMut = origList.get(i);
```
EatingW, line 222: nit: you don't need `multiMut` as a variable, so you can just omit it and say `mutList = origList.get(i).getMutationList();`

src/main/java/com/google/sps/Utility.java
```diff
@@ -207,6 +205,35 @@ public static MultiMutation getMultiMutationAtIndex(List<MultiMutation> multiMut
...
   }
 
+  /**
+   * Returns a set of indices on the original list that related to a given token
+   *
+   * @param tokenName the token name to search for
+   * @param origList the original list of mutations
+   * @return a set of indices, empty if tokenName is null or if token is not changed
+   */
+  public static Set<Integer> getMutationIndicesOfToken(
```
EatingW, line 215: This also seems like a potential optimization opportunity: store a mapping from token to mutation indices?

spec/jstests-spec.js
```diff
@@ -361,29 +361,53 @@ describe("Check correct url params", function () {
...
     const requestParams = requestString.substring(requestString.indexOf("?"));
 
     const constructedUrl = new URLSearchParams(requestParams);
-    expect(constructedUrl.has("nodeName")).toBe(true);
-    expect(constructedUrl.get("nodeName")).toBe("A");
+    expect(constructedUrl.has("nodeNames")).toBe(true); 
+    expect(constructedUrl.get("nodeNames")).toEqual('["A","B"]');
+  });
+
+    it("passes correct nodeName when nodeName has a comma separated value and spaces", function () {
```
EatingW, line 395: nit: indent
Also just thought of another case. What happens for something like "A,,,"? Seems like the frontend would send ["A", "", "", ""] and the server ignores empty strings?

haoyxc: ah I added a filter to check the length > 0!

src/main/java/com/google/sps/DataServlet.java
haoyxc: even if something in nodeNames was not in the graph, would you still want mutations that pertain to it?

IshaniSanturkar: I guess this was just to make sure that if a node in nodeNames was deleted but then added later on (or just in general added after the current step) the tool would still show it.

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -66,6 +66,8 @@
   // to the list [0, mutList.size() - 1].
   HashMap<String, List<Integer>> mutationIndicesMap = new HashMap<>();
 
+  HashMap<String, Set<Integer>> tokenIndicesMap = new HashMap<>();
```
EatingW, line 69: nit: comment on what this mapping is for

haoyxc: think this was taken care of in optimization, my bad!

PR #48, opened 2020-07-29
import individual functions from Utility

haoyxc:
Import Utility.someFunction rather than having to call the function with Utility.someFunction every time. Now they're called with just someFunction.

### DIFF
--- a/src/main/java/com/google/sps/GraphGenerationTest.java
+++ b/src/main/java/com/google/sps/GraphGenerationTest.java
@@ -69,7 +69,7 @@ public final class GraphGenerationTest {
    */
   @Test
   public void correctProtoToGraphNoData() {
-    GraphNode graphNode = Utility.protoNodeToGraphNode(nodeA.build());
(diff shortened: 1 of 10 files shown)


src/main/java/com/google/sps/GraphGenerationTest.java
```diff
@@ -111,7 +113,7 @@ public void correctProtoToGraphWithDataAndMetadata() {
         Struct.newBuilder().putFields("row", rowValue).putFields("column", colValue).build();
     nodeA.setMetadata(metadata);
 
-    GraphNode graphNode = Utility.protoNodeToGraphNode(nodeA.build());
+    GraphNode graphNode = protoNodeToGraphNode(nodeA.build());
     List<String> tokenList = graphNode.tokenList();
     Struct generatedMetadata = graphNode.metadata();
```
IshaniSanturkar, line 118: I noticed in the `correctProtoToGraphNoData` function you still reference `Utility.protoNodeToGraphNode(nodeA.build());`. Accident?

PR #49, opened 2020-07-29
change style & color scheme of page

haoyxc:
Change the layout of page and add colors. Mostly just wanted to put this here to get some concrete feedback!!
LATEST UPDATE:
image

Hover over a button:
Screenshot 2020-07-29 at 4 34 49 PM

### DIFF
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -10,38 +10,41 @@
 </head>
 
 <body id="content" onload="graph.initializeSlider(); graph.generateGraph()">
-  <h1>
+  <h2>
     Graph Visualization Tool
-  </h1>
-  <div id="search-box">
-    <label for="layers">Radius from filtered node(s) (0-20, default is 3):</label>
-    <input type="number" id="num-layers" name="layers" min="0" max="20" value="3">
-    <label for="search-name">Name of node(s) to filter for (If you enter multiple, please separate with commas!) </label>
-    <input type="text" id="node-name-filter" name="search-name" placeholder="nodeName1,nodeName2"></input>
-    <label for="token-name">Name of token to filter for</label>
-    <input type="text" id="token-name-filter" name="token-name" placeholder="Token Name"></input>
-    <button onclick="graph.generateGraph()">Get Graph</button>
-    <div>
-      <label>Zoom on Node in Graph: </label>
-      <input type="text" id="node-search" placeholder="Node Name"></input>
-      <button id="search-button">Find Node</button>
-      <label id="node-error" class="error-text"></label>
+  </h2>
+  <div id="input-boxes">
+    <div id="search-box" class="input-child">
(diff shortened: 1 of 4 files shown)


src/main/webapp/index.html
```diff
@@ -10,38 +10,43 @@
...
-      <label>Zoom on Node in Graph: </label>
-      <input type="text" id="node-search" placeholder="Node Name"></input>
-      <button id="search-button">Find Node</button>
-      <label id="node-error" class="error-text"></label>
+  </h2>
+  <div id="input-boxes">
+    <!-- Box for filtering the graph -->
+    <div id="search-box" class="input-child">
+      <h4 class="search-title">Filter the Graph</h4>
+      <label for="layers">Radius from filtered node(s) (0-20, default is 3):</label>
```
IshaniSanturkar, line 20: Do you think we should rethink this 20 thing now?

src/main/webapp/style.scss
```diff
@@ -4,16 +4,47 @@
...
 }
 
+.search-title {
+  margin-bottom: 20px;
+}
+
+button {
+  background-color: var(--gray);
+  color: black;
+    font-size: 16px;
```
IshaniSanturkar, line 26: Fix indentation here.

IshaniSanturkar: Also `outline: none` removes the weird black outline we see when the button is clicked.

src/main/webapp/index.html
haoyxc: what do you think it should be?

IshaniSanturkar: Maybe the actual max depth? But then we would have to add a backend parameter.

src/main/webapp/index.html
```diff
@@ -10,38 +10,43 @@
...
-    <button id="reset" class="reset">Reset Graph</button>
-    <label id="num-selected"></label>
-    <br />
-    <div id="num-mutation-display">
-      <p id="graph-number-text">Graph </p>
-      <input type="number" id="graph-number" min="0" max="0" value="0"
-        onchange="graph.readGraphNumberInput(); graph.generateGraph()">
-      <p id="total-mutation-number-text">out of 0</p>
+        <label>Zoom on Token in Graph: </label>
+        <input type="text" id="token-search" placeholder="Token Name"></input>
```
IshaniSanturkar, line 44: This is kind of a general question, but what is the relationship between the two zooms? We can't do them both at the same time right?

src/main/webapp/style.scss
```diff
@@ -4,30 +4,67 @@
...
+
+button {
+  background-color: var(--gray);
+  color: black;
+  font-size: 16px;
+  border-radius: 10px;
+  margin: 5px;
+  padding: 5px;
+  border-style: solid;
+  border-width: 0;
```
IshaniSanturkar, line 35: Just curious, what is the purpose of border styling if the width is 0 :)

PR #50, opened 2020-07-30
Change previous button logic to generate graph from current graph rather than original graph if more efficient

IshaniSanturkar:
In this PR, I:
- changed the logic so that if it requires fewer mutations to go from the current graph to the next graph (where the next graph is earlier in time than this one) than from the initial graph to the next graph, we do so.
- added backend tests for the new functions.


src/main/java/com/google/sps/DataServlet.java
```diff
@@ -52,6 +52,11 @@
   private DataGraph currDataGraph = null;
   private DataGraph originalDataGraph = null;
   private List<MultiMutation> mutList = null;
+  // We store a builder so that we can modify the contained mutation objects
+  // to be non-redundant. For example if we find that a mutation adds duplicate
```
haoyxc, line 64: nit: comma after For example

src/main/java/com/google/sps/DataGraph.java
```diff
@@ -309,27 +315,49 @@ public String mutateGraph(Mutation mut) {
...
       // Update the map
       for (String tokenName : tokenNames) {
         addNodeToToken(tokenName, node.name());
       }
     } else if (tokenMutType == TokenMutation.Type.DELETE_TOKEN) {
       tokenList.removeAll(tokenNames);
+
+      if (tokenList.size() != existingTokens.size() - tokenNames.size()) {
+        // Remove tokens that this mutation deletes that don't exist in the node
+        tokenNames = new ArrayList<>(tokenNames);
```
haoyxc, line 357: same thing here -- but could be missing something

src/main/java/com/google/sps/DataGraph.java
```diff
@@ -309,27 +315,49 @@ public String mutateGraph(Mutation mut) {
...
 
     TokenMutation.Type tokenMutType = tokenMut.getType();
     if (tokenMutType == TokenMutation.Type.ADD_TOKEN) {
       tokenList.addAll(tokenNames);
 
+      if (tokenList.size() != existingTokens.size() + tokenNames.size()) {
+        // Remove tokens that this mutation adds that already exist in the node
+        // we must reinstantiate tokenNames because the one retrieved from
+        // the token mutation is unmodifiable
+        tokenNames = new ArrayList<>(tokenNames);
```
haoyxc, line 343: is there a reason to set tokenNames to itself/ what exactly is the point of this line?

src/main/java/com/google/sps/DataGraph.java
```diff
@@ -309,27 +315,49 @@ public String mutateGraph(Mutation mut) {
...
+  private GraphNode changeNodeToken(GraphNode node, TokenMutation.Builder tokenMut) {
 
     // List of tokens to add/remove from the existing list
     List<String> tokenNames = tokenMut.getTokenNameList();
     // The existing list of tokens in the node
-    List<String> tokenList = new ArrayList<>();
-    tokenList.addAll(node.tokenList());
+    List<String> existingTokens = node.tokenList();
+    // The modified list of tokens of the node
+    Set<String> tokenList = new HashSet<>();
```
haoyxc, line 332: could we change the variable name to tokenSet instead of tokenList since it's a set?

src/main/java/com/google/sps/Utility.java
```diff
@@ -108,55 +110,66 @@ public static String graphToJson(
    * @param original the original graph
    * @param curr the current (most recently-requested) graph (requires that original != curr)
    * @param mutationNum number of mutations to apply
-   * @param multiMutList multi-mutation list
+   * @param multiMutList multi-mutation list builder
```
haoyxc, line 113: maybe include here that the param could be modified?

src/main/java/com/google/sps/Utility.java
```diff
@@ -108,55 +110,66 @@ public static String graphToJson(
...
+
     if (mutationNum < -1) {
       return null;
     } else if (mutationNum > multiMutList.size()) {
       mutationNum = multiMutList.size() - 1;
     }
+    // If the requested graph is sequentially before the current graph but is closer to
+    // the initial graph than to the current graph, go forward from the initial graph rather
+    // than going back from the current graph
+    if (curr.numMutations() > mutationNum && (curr.numMutations() - mutationNum) > mutationNum) {
```
haoyxc, line 135: maybe put this in the else block on line 158 and just keep the `(curr.numMutations() - mutationNum) > mutationNum` part?

src/main/webapp/WEB-INF/mutation.textproto
```diff
@@ -88,11 +89,22 @@ mutation {
...
     type: DELETE_TOKEN
     token_name: "d.js"
+    token_name: "x.js"
   }
 }
 reason: "Changing tokens of node B"
 }
 mutation {
+  mutation {
+  type: CHANGE_TOKEN
```
haoyxc, line 99: nit: indent?

src/main/java/com/google/sps/DataGraph.java
IshaniSanturkar: I'll change it here, but I think the name in the proto is still tokenList :(.

src/main/java/com/google/sps/Utility.java
```diff
@@ -101,71 +106,82 @@ public static String graphToJson(
...
+      curr = original.getCopy();
+    }
     if (curr.numMutations() <= mutationNum) { // going forward
       for (int i = curr.numMutations() + 1; i <= mutationNum; i++) {
         // Mutate graph operates in place
         MultiMutation multiMut = multiMutList.get(i);
         List<Mutation> mutations = multiMut.getMutationList();
+        // Use this to store the multi mutation without any redundant mutation
+        // information (for eg. duplicate tokens to add)
+        MultiMutation.Builder trimmedMultiMut = MultiMutation.newBuilder();
```
EatingW, line 148: Right now each mutation would create a trimmedMultiMut and update the mutationsList, but only one update is needed for every mutation. One potential optimization is to keep track of which mutation is already updated to the "trimmed" version to avoid always creating a builder and updating the trimmedMultiMut.

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -179,118 +193,162 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
       filteredDiff = filterMultiMutationByNodes(diff, Sets.union(truncatedGraphNodeNames, queried));
     }
+
+    /*
+     ***********************
+     * Error Handling
+     ***********************
+     */
+
     // We set the headers in the following 4 scenarios:
```
EatingW, line 304: nit:This comment should be moved to the above "Error Handling" section

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -179,118 +193,162 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-    // The searched node is not in the graph and is never mutated
+    // truncatedGraph.nodes().size() == 0 means something was queried but wasn't found
```
EatingW, line 305: nit: I don't think you need to copy the actual code here. You can inline the comment with the code:
```
if (truncatedGraph.nodes().size() == 0 // something was queried but wasn't found in the graph
    && filteredMutationIndices.size() == 0) { // the searched object is never mutated
```

src/main/java/com/google/sps/DataServlet.java
```diff
@@ -179,118 +193,162 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-    // when the searched node is deleted for example. The diff being non-empty
-    // means that there is some mutation pertaining to the searched node to show
+    // truncatedGraph.nodes().size() == 0 means something was queried but wasn't found
+    // in the graph
+    // filteredMutationIndices.size() != 0 means the searched object is mutated at some
+    // point
+    // filteredDiff == null || filteredDiff.getMutationList().size() == 0 means that none
+    // of the searched objects are mutated here. This exists to prevent this message from
+    // being emitted when the searched object is deleted in this graph so it doesn't exist
+    // but is still mutated
```
EatingW, line 321: nit: same here, but you might need to shorten the comments to fit within one line, something like:
```
if (truncatedGraph.nodes().size() == 0 // searched object not found in the graph
        && filteredMutationIndices.size() != 0 // ... and mutated at some point
        && (filteredDiff == null || filteredDiff.getMutationList().size() == 0)) { // ... but not mutated here
```

src/main/java/com/google/sps/Utility.java
IshaniSanturkar: @EatingW could we look into it after the presentation :)

spec/jstests-spec.js
```diff
@@ -632,24 +649,57 @@ describe("Token search", function () {
...
+
+    nextButton.click();
+    expect(otherNode.hasClass("highlighted-node")).toBe(true);
+    expect(otherNode.hasClass("non-highlighted-node")).toBe(false);
+
+    expect(result.hasClass("highlighted-node")).toBe(false);
+    expect(result.hasClass("non-highlighted-node")).toBe(true);
+    expect(prevButton.disabled).toBe(false);
+    expect(nextButton.disabled).toBe(true);
+
```
haoyxc, line 686: nit: extra line

spec/jstests-spec.js
```diff
@@ -841,7 +891,8 @@ describe("Showing and hiding tooltips when checkbox is clicked", function () {
     <button id="clear-log">Clear Log</button>
     <button id="reset"></button>
     <button id="search-token-button"></button>
-    <input type="number" id="highlight-number">`;
+    <input type="number" id="highlight-number">
+    <button id="gen-graph" onclick="graph.generateGraph()">Get Graph</button>`;
```
haoyxc, line 897: also extra line

PR #51, opened 2020-08-01
Improve overall readability and usability

connorbrett:
image

Improvements/features in this PR:
- node names move from label to inside tooltip
- clear logs button
- navigate through nodes that contain specific token

### DIFF
--- a/spec/jstests-spec.js
+++ b/spec/jstests-spec.js
@@ -2,9 +2,10 @@
 import {
   initializeNumMutations, setMutationIndexList, setCurrMutationNum, setCurrMutationIndex,
   initializeTippy, generateGraph, getUrl, navigateGraph, currMutationNum, currMutationIndex,
-  numMutations, updateButtons, searchAndHighlight, highlightDiff, initializeReasonTooltip, getGraphDisplay,
-  getClosestIndices, initializeSlider, resetMutationSlider, mutationNumSlider, setMutationSliderValue, 
-  readGraphNumberInput, updateGraphNumInput, setMaxNumMutations, searchNode, searchToken, clearLog
+  numMutations, updateButtons, searchAndHighlight, highlightDiff, initializeReasonTooltip, 
+  getGraphDisplay, getClosestIndices, initializeSlider, resetMutationSlider, mutationNumSlider, 
+  setMutationSliderValue, readGraphNumberInput, updateGraphNumInput, setMaxNumMutations, 
+  searchNode, searchToken, clearLogs
 }
   from "../src/main/webapp/script.js";
 
@@ -720,6 +721,19 @@ describe("Initializing mutation reason tooltips", function () {
   });
 });
 
+/** Clearing the logs works properly */
+describe("Clearing the logs", function () {
+  it("correctly clears the log list", function () {
+    document.body.innerHTML = `
+    <ul id="log-list">
+        <li class="log-msg">hi </li>
+      </ul>`;
+      const lst = document.getElementById('log-list');
+      clearLogs();
+      expect(lst.firstChild).toBe(null); 
+  })
+})
+
 describe("Showing and hiding tooltips when checkbox is clicked", function () {
   it("correctly shows/hides tooltips when checkbox is checked/unchecked", function () {
     document.body.innerHTML = `
@@ -1034,7 +1048,7 @@ describe("Testing clear log functionality", function () {
       </div>`;
     logList = document.getElementById("log-list");
     clearLogButton = document.getElementById("clear-log");
-    clearLogButton.onclick = function () { clearLog() };
+    clearLogButton.onclick = function () { clearLogs() };
   });
 
   it("clears the logs when they contain messages", function () {
--- a/spec/jstests-spec.js
+++ b/spec/jstests-spec.js
@@ -296,12 +296,12 @@ describe("Pressing next and previous buttons associated with a graph", function 
     expect(prevButton.disabled).toBe(true);
   });
 
-  it("Correctly doesn't change anything when there aren't any mutations", function() {
+  it("Correctly doesn't change anything when there aren't any mutations", function () {
     document.body.appendChild(numDisplay);
     initializeNumMutations(-1);
     setCurrMutationNum(10);
-   
-   // Nothing changes with a negative numMutations
+
+    // Nothing changes with a negative numMutations
     navigateGraph(-1);
     expect(currMutationNum).toBe(10);
 
@@ -309,7 +309,7 @@ describe("Pressing next and previous buttons associated with a graph", function 
     expect(currMutationNum).toBe(10);
   });
 
-  it("Correctly navigates forward when index is a decimal", function() {
+  it("Correctly navigates forward when index is a decimal", function () {
     document.body.appendChild(numDisplay);
     setCurrMutationIndex(.5);
     setCurrMutationNum(0);
@@ -319,7 +319,7 @@ describe("Pressing next and previous buttons associated with a graph", function 
     expect(currMutationNum).toBe(1);
   });
 
-  it("correctly navigates backward when index is a decimal", function() {
+  it("correctly navigates backward when index is a decimal", function () {
     document.body.appendChild(numDisplay);
     setCurrMutationIndex(.5);
     setCurrMutationNum(0);
@@ -828,9 +828,8 @@ describe("Testing graph input functionality", function () {
   let totalMutNumberText;
 
   beforeEach(function () {
-    document.body.innerHTML = 
-    `
-      <div id="num-mutation-display">
+    document.body.innerHTML =
+      `<div id="num-mutation-display">
         <p id="graph-number-text">Graph </p>
         <input type="number" id="graph-number"  min="0" max="0" value="0">
         <p id="total-mutation-number-text">out of 0</p>
@@ -843,12 +842,12 @@ describe("Testing graph input functionality", function () {
 
   it("correctly updates graph number input when mutation list is empty", function () {
     setMutationIndexList([]);
-    initializeNumMutations(0);
     setCurrMutationNum(-1);
     setCurrMutationIndex(-1);
     setMaxNumMutations(0);
 
     updateGraphNumInput();
+    // Just the initial graph
     expect(graphNumberInput.value).toBe("0");
     expect(graphNumberInput.min).toBe("0");
     expect(graphNumberInput.max).toBe("0");
@@ -862,12 +861,12 @@ describe("Testing graph input functionality", function () {
 
   it("correctly updates graph number input when the mutation index list is non-empty", function () {
     setMutationIndexList([1, 5, 7, 11]);
-    initializeNumMutations(4);
     setCurrMutationNum(1);
     setCurrMutationIndex(0);
     setMaxNumMutations(20);
 
     updateGraphNumInput();
+    // Current graph is the graph after mutation 1
     expect(graphNumberInput.value).toBe("2");
     expect(graphNumberInput.min).toBe("0");
     expect(graphNumberInput.max).toBe("20");
--- a/spec/jstests-spec.js
+++ b/spec/jstests-spec.js
@@ -822,7 +822,6 @@ describe("Testing slider functionality", function () {
   });
 });
 
-
 describe("Testing graph input functionality", function () {
   let graphNumberInput;
   let totalMutNumberText;
--- a/spec/jstests-spec.js
+++ b/spec/jstests-spec.js
@@ -562,8 +562,8 @@ describe("Token search", function () {
   let cy;
 
   beforeEach(function () {
-    document.body.innerHTML =
-      `<div id="cy"></div>
+    document.body.innerHTML = `
+      <div id="cy"></div>
       <div id="zoom-box" class="input-child">
         <h4 class="search-title">Zoom In</h4>
           <label>Zoom on Node in Graph: </label>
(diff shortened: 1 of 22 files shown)


src/main/webapp/script.js
```diff
@@ -708,46 +756,50 @@ function searchToken(cy, query) {
...
+    });
   });
-  document.getElementById('num-selected').innerText = "Number of nodes selected: 0";
+
+  if(resetInputs === true) {
+    document.getElementById("node-search").value = '';
+    document.getElementById("token-search").value = '';
+    document.getElementById('num-selected').innerText = "out of 0";
+    document.getElementById('highlight-number').value = 0;
+    document.getElementById('highlight-number').max = 0;
```
IshaniSanturkar, line 799: Also reset min.

src/main/webapp/script.js
```diff
@@ -348,11 +361,23 @@ function getGraphDisplay(graphNodes, graphEdges, mutList, reason, queriedNodes)
...
-  document.getElementById('reset').onclick = function(){ resetElements(cy) };
+  document.getElementById('reset').onclick = function () { resetElements(cy, true) };
+
+  document.getElementById('search-node-button').onclick = function () { searchAndHighlight(cy, "node", searchNode) };
 
-  document.getElementById('search-button').onclick = function() { searchAndHighlight(cy, "node", searchNode) };
+  document.getElementById('search-token-button').onclick = function () { searchAndHighlight(cy, "token", searchToken) };
 
-  document.getElementById('search-token-button').onclick = function() { searchAndHighlight(cy, "token", searchToken) };
+  // document.getElementById('highlight-number').onchange = function() {
```
EatingW, line 370: delete?

src/main/webapp/script.js
```diff
@@ -696,9 +719,46 @@ function searchToken(cy, query) {
...
-    return target;
+    document.getElementById('num-selected').innerText = "out of " + target.length;
+    document.getElementById('highlight-number').min = 1;
+    document.getElementById('highlight-number').value = 1;
+    document.getElementById('highlight-number').max = target.length;
+    document.getElementById('highlight-number').disabled = false;
+    document.getElementById('highlight-number').onchange = function () {
+      updateHighlightedToken(cy, target, document.getElementById('highlight-number').value - 1);
+    };
+    document.getElementById('prevnode').onclick = function () {
```
EatingW, line 732: nice is this based on my FR during the daily sync? I think the buttons should probably only be displayed if the target.length > 1 because there's no point having such buttons when there is only one node to highlight.

spec/jstests-spec.js
```diff
@@ -427,79 +439,150 @@ describe("Check correct url params", function () {
...
+describe("Token search", function () {
+  let logList;
   let query;
   let cy;
 
-  beforeEach(function() {
-    document.body.innerHTML = `<div id="cy"></div>`;
+  beforeEach(function () {
+    document.body.innerHTML =
+      `<div id="cy"></div>
```
EatingW, line 566: nit: indentation seems a bit off here.

src/main/webapp/script.js
connorbrett: yes it is, should they be removed completely (display: none) or just disabled? also should they be removed/disabled when token search is not being used? ie. node search or filtering was just executed

IshaniSanturkar: I'm working on the disabled part. I'll push in a few.

haoyxc: i think it should be removed? but either could work what do you think?

src/main/webapp/index.html
```diff
@@ -89,6 +92,7 @@ <h4 class="search-title">Zoom In</h4>
       </div>
     </div>
     <!-- COLUMN 2: includes messages -->
+
```
haoyxc, line 95: did you mean to add an empty line?

src/main/webapp/script.js
```diff
@@ -348,11 +360,23 @@ function getGraphDisplay(graphNodes, graphEdges, mutList, reason, queriedNodes)
...
+  //   updateHighlightedToken(cy, document.getElementById('highlight-number').value); 
+  // };
+
+  // document.getElementById('prevnode').onclick = function () { 
+  //   updateHighlightedToken(cy, document.getElementById('highlight-number').value - 1);
+  // };
+
+  // document.getElementById('nextnode').onclick = function () { 
+  //   updateHighlightedToken(cy, document.getElementById('highlight-number').value + 1);
+  // };
```
haoyxc, line 379: delete?

src/main/webapp/script.js
```diff
@@ -532,6 +560,57 @@ function highlightDiff(cy, mutList, reason = "") {
   };
 }
 
+function indicateChangedTokens(node, tokenMut) {
+  const type = tokenMut["type_"] || -1;
+  const tokens = tokenMut["tokenName_"];
+
+  if(!type || !tokens) {
+
```
haoyxc, line 568: maybe return?

src/main/webapp/script.js
```diff
@@ -696,58 +773,115 @@ function searchToken(cy, query) {
...
+  } else {
+    document.getElementById('prevnode').disabled = false;
+  }
+  if (num >= nodesWithToken.length - 1) {
+    num = nodesWithToken.length - 1;
+    document.getElementById('nextnode').disabled = true;
+  } else {
+    document.getElementById('nextnode').disabled = false;
+  }
+  nodesWithToken.toggleClass('highlighted-node', false);
```
haoyxc, line 823: could you document this a liiiiittle more?

spec/jstests-spec.js
```diff
@@ -427,79 +450,156 @@ describe("Check correct url params", function () {
...
     query.value = "A";
     const result = searchAndHighlight(cy, "node", searchNode);
-
     // should not display error message
-    expect(nodeError.innerText).toBe("");
+    expect(logList.textContent.trim()).toBe("");
     expect(result.id()).toBe("A");
-    expect(result.style("border-width")).toBe(expectedBorderWidth);
+    expect(result.hasClass("highlighted-node")).toBe(true);
+    expect(result.hasClass("non-highlighted-node")).toBe(false);
```
haoyxc, line 551: check the other nodes don't have the class like you did below for this as well?

src/main/webapp/script.js
```diff
@@ -696,58 +773,115 @@ function searchToken(cy, query) {
...
+    cy.edges().forEach(edge => {
+      edge.toggleClass('highlighted-edge', false);
+    });
   });
-  document.getElementById('num-selected').innerText = "Number of nodes selected: 0";
+
+  // Set all logs to be black
+  resetRecentLogs();
+
+  if (resetInputs === true) {
```
haoyxc, line 873: why do you need the === true?

src/main/webapp/script.js
```diff
@@ -742,22 +830,40 @@ function searchToken(cy, query) {
...
+  // Also, if we are at the last node, disable the next button
+  // so the user can't go forward
+  if (num >= nodesWithToken.length - 1) {
     num = nodesWithToken.length - 1;
+    document.getElementById('nextnode').disabled = true;
+  } else {
+    document.getElementById('nextnode').disabled = false;
   }
+
+  // Remove highlight from all nodes and associated edges
```
EatingW, line 859: This removes the highlighting from the other searched nodes while focusing only on one of them, right? I actually think it's useful to highlight all of the searched nodes (if I choose to drag the graph around to see them), and clicking "prevnode" or "nextnode" should only help me zoom in on the node for clarity, but not remove the highlighting.

src/main/webapp/script.js
```diff
@@ -557,6 +571,75 @@ function highlightDiff(cy, mutList, reason = "") {
...
+    return;
+  }
+
+  // Get the div that contains the tooltip content
+  let tipContent = node.tip.popperChildren.content.firstChild;
+  switch (type) {
+    case 1:
+      // add tokens
+      // If this node already has an "Added Token" list, get it
+      let addedList = tipContent.querySelector(`#${node.data().id}-added`);
```
EatingW, line 595: I wonder if it makes more sense to also reset the addedList and deletedList, i.e. if this mutation didn't mutate any tokens, the tooltip should also have empty "added tokens" and "deleted tokens" sections. What do you think?

src/main/webapp/script.js
```diff
@@ -220,7 +219,7 @@ function disableInputs() {
 function getUrl() {
   const depthElem = document.getElementById('num-layers');
   const nodeNames = document.getElementById('node-name-filter') ? document.getElementById('node-name-filter').value || "" : "";
```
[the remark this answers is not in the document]
IshaniSanturkar: I think we do that when we create `nodeNamesArray`?

src/main/webapp/script.js
IshaniSanturkar: Should we still only show the edges of the zoomed-in node in dashed form?

src/main/webapp/script.js
IshaniSanturkar: Since nodes which were mutated are highlighted, do you think adding empty lists would be useful?

src/main/webapp/script.js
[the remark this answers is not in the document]
EatingW: Ah right!

src/main/webapp/script.js
EatingW: yeah, my comment was mainly about node-highlighting

src/main/webapp/script.js
EatingW: What I meant was, if node A's token were mutated at mutation N, I would expect to see the  "Added tokens" and "Deleted tokens" in the tooltip. But on mutation N+1, if I click on node A again I shouldn't see any added or deleted tokens.

EatingW: Actually I think I'm mistaken here. I got confused by "If this node already has an "Added Token" list". The tippy is always newly created for each node, right? So shouldn't it always only contain a token list section before this function is called? Why would the tippy already contain an "Added Token" list?

EatingW: Actually I think I'm mistaken here. I got confused by "If this node already has an "Added Token" list". The tippy is always newly created for each node, right? So shouldn't it always only contain a token list section before this function is called? Why would the tippy already contain an "Added Token" list?

IshaniSanturkar: I was just considering the case where there are two change token mutations for a node in the multi-mutation diff. I didn't want it to create separate "Added token" lists in that case. Do you think that situation is possible?

EatingW: Ah ok, from my exploring of data that's definitely possible :)

src/main/webapp/index.html
```diff
@@ -35,18 +34,21 @@ <h4 class="search-title">Filter the Graph</h4>
     <!-- Box for zooming in -->
     <div id="zoom-box" class="input-child">
       <h4 class="search-title">Zoom In</h4>
-        <label>Zoom on Node in Graph: </label>
-        <input type="text" id="node-search" placeholder="Node Name"></input>
-        <button id="search-button">Find Node</button>
-        <label id="node-error" class="error-text"></label>
+      <label>Zoom on Node in Graph: </label>
```
haoyxc, line 37: can you add something about the behavior regarding zooming in on a node and a token? (is that allowed?)

src/main/webapp/index.html
```diff
@@ -35,18 +34,21 @@ <h4 class="search-title">Filter the Graph</h4>
...
-      <button id="reset" class="reset">Reset Zoom Effects</button>
-      <label id="num-selected"></label>
+      <label>Zoom on Token in Graph: </label>
+      <input type="text" id="token-search" placeholder="Token Name"></input>
+      <button id="search-token-button">Find Token</button>
+      <div>
+        <label for="highlight-number">Highlighted Token: </label>
+        <input type="number" id="highlight-number" min="0" value="0" max="0" disabled=true>
+        <label id="num-selected">out of 0</label>
+        <button id="prevnode" class="material-icons" disabled=true>navigate_before</button>
```
haoyxc, line 48: consider hiding this when no token is searched & when there's one searched? it can be a bit confusing for the user

src/main/webapp/style.scss
```diff
@@ -178,6 +188,19 @@ body {
...
 input {
   padding: 5px;
 }
+
+.deletedtoken {
+  color: red;
+}
+
+.addedtoken {
+  color: green;
```
haoyxc, line 205: nit: consider adding green to the root & using it (also can change the name of the variable in root for red and use it in deletedtoken

src/main/webapp/index.html
connorbrett: atm they are disabled in those scenarios, do you think not displaying them at all would be better?

src/main/webapp/script.js
```diff
@@ -532,6 +580,75 @@ function highlightDiff(cy, mutList, reason = "") {
...
+    return;
+  }
+
+  // Get the div that contains the tooltip content
+  let tipContent = node.tip.popperChildren.content.firstChild;
+  switch (type) {
+    case 1:
+      // add tokens
+      // If this node already has an "Added Token" list, get it
+      let addedList = tipContent.querySelector(`#${node.data().id}-added`);
```
EatingW, line 604: Based on the bug we discussed via chat, I verified that doing `tipContent.querySelector(`#${CSS.escape(node.data().id)}-added`);` works here.

Same for the selector of `-deleted`

IshaniSanturkar: Thank you :)

src/main/webapp/script.js
```diff
@@ -505,7 +546,7 @@ function highlightDiff(cy, mutList, reason = "") {
         modifiedObj = cy.$id(`edge${startNode}${endNode}`);
         modifiedObj.style('line-color', colorScheme["deletedObjectColor"]);
         modifiedObj.style('target-arrow-color', colorScheme["deletedObjectColor"]);
-        modifiedObj.style('opacity', opacityScheme["deletedObjectOpacity"]);
+        modifiedObj.style('opacity', opacityScheme["translucentObjectOpacity"]);
```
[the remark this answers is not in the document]
haoyxc: the blurring works with zooming in on nodes though!!

PR #52, opened 2020-08-04
Add checkbox to restrict mutations only to nodes & tokens filtered for

haoyxc:
Add option to restrict the mutations to only the nodes/tokens filtered for. Not checked by default. Also added frontend tests to check the url paramter

image


spec/jstests-spec.js
```diff
@@ -450,6 +450,27 @@ describe("Check correct url params", function () {
...
+    expect(constructedUrl.has("restrict")).toBe(true);
+    expect(constructedUrl.get("restrict")).toBe("false");
+
+    limitMutCheckbox.click();
+    requestString = getUrl();
+    requestParams = requestString.substring(requestString.indexOf("?"));
+    constructedUrl = new URLSearchParams(requestParams);
+    expect(constructedUrl.has("restrict")).toBe(true);
+    expect(constructedUrl.get("restrict")).toBe("true");
+  })
```
IshaniSanturkar, line 471: Nit: semicolons here and the line below.

